PR #3, opened 2020-05-05
A "Hello World" to start with our app

denoribi:
Here is our short "Hello World" Program/App. I used some features of Ionic and Angular, like buttons, input and routing to create a "Homepage"(=starting page in our app) and to generate a little interaction with the User. Unfortunately i needed to delete the organization of the imports, because there is a bug in vscode on mac.


frontend/src/app/home/home.page.ts
```diff
@@ -1,8 +1,25 @@
...
 })
-export class HomePage {}
+export class HomePage {
+  checkOutForm = this.formBuilder.group({ firstName: [''], lastName: [''] });
+
+  constructor(
+    private formBuilder: FormBuilder,
+    private navController: NavController
+  ) {}
+  myS(): void {
```
MBuchalik, line 17: For future reference: You might want to assign better names to your methods, and maybe add some comments telling what they do. This is no problem at the moment, since this is really just a "hello world" application.

frontend/src/app/home/home.page.html
```diff
@@ -1,28 +1,48 @@
...
     </p>
+    <hr />
+    <form [formGroup]="checkOutForm">
+      <ion-item>
+        <ion-label>Name</ion-label>
+        <ion-input
+          required
+          type="text"
+          formControlName="firstName"
+          style="color: gray;"
```
MBuchalik, line 31: For future reference: You should apply inline styles as rarely as possible. Also, for Ionic components, in most cases you need to use CSS Variables (you can find them in the Ionic Docs, for instance https://ionicframework.com/docs/api/input#css-custom-properties ). I haven't tested it yet, but I believe that `style="color: gray;"` has no effect here.

This is no problem at the moment, since this is really just a "hello world" application.

MBuchalik:
LGTM 👍 

I added comments regarding the code style. You don't need to fix this now since it is just a basic Hello World app. But in the future, I will probably complain more about stuff like this 😉

frontend/src/app/home/home.page.ts
denoribi: Yeah, thought this will be deleted anyway, so i didn't had an eye on it, but for the future i will get better names  :)

frontend/src/app/home/home.page.html
denoribi: Okay, i will have a look on it, thanks :)

PR #14, opened 2020-05-12
Added 3 linter rules 

Bharathwaj-Ravi-1995:
I have added 3 linter rules such as 
- no-conditional-assignment
- no-sparse-arrays
- no-boolean-literal-compare
Which can restrict these common code mistakes.


MBuchalik:
Good ideas! Tested it locally and it works fine 👍 Definitely prevents some really common issues.

PR #21, opened 2020-05-18
Add Roboto and a more simple logo

### LINKED ISSUE #2: Include a local copy of Roboto
If I am not mistaken, Ionic doesn't come with a local version of Roboto. This is a problem when developing in the browser since it will fall back to another font that most likely doesn't include all font-weights included in Roboto.

MBuchalik:
I added a local copy of Roboto. Also, I extracted the "O" from the logo and directly embedded it into the application.

Closes #2


Bharathravi1995 (approved):
Looks fine to me. In future, we should try to avoid inline SVG if it is used in many places for better maintainence.

denoribi:
looks good for me

MBuchalik:
> Looks fine to me. In future, we should try to avoid inline SVG if it is used in many places for better maintainence.

I have inlined the SVG so that we can easily control the colors. For instance, when the device switches to Dark Mode, the logo will now have different colors applied to it. I don't know how we could achieve this with an img-included SVG.

If we wanted to use the Logo in more places, we could refactor it into a separate Angular Component.

PR #23, opened 2020-05-19
Basic information step changes done.

Bharathravi1995:
**This PR Contains following changes** 

- data flow and event flow between parent and child components and vice versa

- Basic information view related changes

- side hamburger menu list related changes

**Some common steps to follow while adding new form step:**

1. Each step content has to be a different component.

2. Whenever new view is added add an object 
`{
      id: <consecutive_number>,
      displayName: <display_name_in_side_menu>,
      selector: <selector_of_the_component>,
      isActive: false,
      isCompleted: false,
    },`
 in _sideMenuList_ property of  forms.page.ts

3. Create an Interface for your fields and import in forms.page.ts and your component.

4. Create a typesafe form groups and control property in forms.page.ts

5. Add switch case entry of your component  with 

```
<your-component
      (callBackNameInYourComponent)="pageProgressStatusCallBack($event)"
      [menuObjectNameInYourComponent]="currentMenu"
      [formsGroupPropertyNameInYourComponent]="yourFormsGroupPropert"
      *ngSwitchCase=<selector>>
</your-component>
```

6. In your component emit an event to parent page method _pageProgressStatusCallBack($event)_ with the information of whether the form is completed by assigning value to _isCompleted_ property in _menuObject_


frontend/src/app/forms/forms-routing.module.ts
```diff
@@ -0,0 +1,27 @@
+/*
+@Author
+Bharathwaj Ravi
+
+Add modifiers under @Modifier
+@Modifiers
+
+@Purpose
+  - This page has the route path configuration in forms module.
+*/
```
MBuchalik, line 10: Please remove this header. It is really uncommon to have a header like this in a TS project. Also, since we are all working on the project, you are not the only Author of this file.

(Also, it is self-exmpanatory that this page contains routes.)

I will not comment on the other files, but please remove these headers there as well.

MBuchalik: The imports should always be the first thing in a file. I should have added a linter rule for that 😄

frontend/src/app/forms/forms.page.html
```diff
@@ -0,0 +1,55 @@
...
+<app-side-menu
+  [sideMenuList]="sideMenuList"
+  (pageProgressStatusCallBack)="onFormStepsNavigation($event)"
+>
+</app-side-menu>
+
+<ion-content class="ion-padding">
+  <ng-container [ngSwitch]="currentMenu.selector">
+    <form-basic-info
+      (pageProgressStatusCallBack)="pageProgressStatusCallBack($event)"
```
MBuchalik, line 37: You might want to consider renaming 
```
(pageProgressStatusCallBack)="pageProgressStatusCallBack($event)"
```
to something like
```
(pageProgressStatusChange)="pageProgressStatusCallBack($event)"
```
or similar. Simply to sound more like an event. But that's up to you.

frontend/src/app/forms/forms.page.scss
```diff
@@ -0,0 +1,18 @@
+.width90 {
+  width: 90%;
+}
+
+.font30 {
+  font-size: 30px;
+}
```
MBuchalik, line 7: It is really uncommon to have class names like this (except for some rare library things). What is the purpose of the element? 
For instance, the ".width90". Isn't this rather a "container"/"main-container"/"main-content"/"wrapper"?

frontend/src/app/forms/forms.page.html
```diff
@@ -0,0 +1,55 @@
...
+            class="ion-float-right font30"
+            autoHide="true"
+            menuClose
+            id="main"
+          >
+            <ion-icon name="list"></ion-icon>
+          </ion-menu-button>
+        </ion-col>
+      </ion-row>
+    </ion-grid>
```
MBuchalik, line 22: You definitely made things way to complicated ;) Just have a look at https://ionicframework.com/docs/api/header

Here, you can see that there is a "slot" you can define for your icons. And the header text can also easily be added. (Just keep in mind that the Android version looks slightly different... But that shouldn't matter at the moment.)

So your HTML would look sth like this (careful! Only a simple "illustration" and not the final solution)

``` HTML

<ion-header>
  <ion-toolbar>
  <ion-buttons slot="start">
      <ion-button
        ><ion-icon name="close" (click)="onClose()"></ion-icon
      ></ion-button>
    </ion-buttons>

    <ion-title> Step {{currentMenu.id}}/ {{totalSteps}}</ion-title>

    <ion-buttons slot="end">
      <ion-button><ion-icon name="list"></ion-icon></ion-button>
    </ion-buttons>
  </ion-toolbar>
  
  <ion-progress-bar value="{{progressPercentage}}"></ion-progress-bar>
</ion-header>
```

frontend/src/app/forms/forms.page.scss
MBuchalik: Additionally: Your content currently looks pretty good on iPhone, but imho does not really follow the ideas I provided for iPad. But we can try to improve it over time. :)

frontend/src/app/forms/forms.page.ts
```diff
@@ -0,0 +1,175 @@
...
+import { FormControl, FormGroup } from './../common/forms/forms';
+
+@Component({
+  selector: 'app-forms',
+  templateUrl: './forms.page.html',
+  styleUrls: ['./forms.page.scss'],
+})
+export class FormsPage implements OnInit, OnDestroy {
+  /*
+  @Usage this array holds the object with   information required for different views.
```
MBuchalik, line 29: I saw it in many places, but will only write it down here: Where is this `@Usage`/`@Whatever` notation from? I have never seen it on any TS project so far.

I would highly recommend adding comments like so:

``` TS
/**
 * This is a comment...
 */
sideMenuList = ...
```


The reason: Having the two asterisk at the beginning (`/**`) will allow us to use the  Intellisense (or whatever it is called) feature of VSCode. So, whenever you hover "this.sideMenuList", you will get a popup with your message.

Please update this in all places where you added a comment right above a method or attribute or class. The only time where you don't add two asterisk to a comment is when it doesn't belong to and class/attribute/method/... in particular.

frontend/src/app/forms/forms.page.ts
```diff
@@ -0,0 +1,175 @@
...
+  */
+  ngOnDestroy(): void {
+    for (const subscription of this.subscriptions) {
+      subscription.unsubscribe();
+    }
+    this.basicInfoObj = new FormGroup<BasicInfo>({
+      firstName: new FormControl(''),
+      lastName: new FormControl(''),
+      salutation: new FormControl('mr'),
+    });
```
MBuchalik, line 131: What does this piece do? When the Component is destroyed, we don't need the FormGroup anymore anyways. So why reset it?

frontend/src/app/forms/forms.page.ts
```diff
@@ -0,0 +1,175 @@
...
+  @Usage In this methos progress values are updated.
+  */
+  pageProgressStatusCallBack(event: {
+    id: number;
+    displayName: string;
+    isCompleted: boolean;
+    isActive: boolean;
+    selector: string;
+  }): void {
+    console.log(event);
```
MBuchalik, line 170: Maybe, remove the console log?

frontend/src/app/forms/side-menu/side-menu.component.scss
```diff
@@ -0,0 +1,3 @@
+.font25 {
```
MBuchalik, line 1: The same as for the forms.page.scss

frontend/src/app/forms/steps/basic-info/basic-info.component.ts
```diff
@@ -0,0 +1,71 @@
...
+  */
+  salutationsArray = [
+    { value: 'mr', displayName: 'Mr' },
+    { value: 'mrs', displayName: 'Mrs' },
+    { value: 'ms', displayName: 'Ms' },
+  ];
+
+  constructor() {}
+
+  ngOnInit(): void {}
```
MBuchalik, line 54: Empty --> Can be removed.

frontend/src/app/forms/steps/basic-info/basic-info.component.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  /*
+  @Usage It handles the on change event for the fields and send emit call back event to parent.
+  */
+  onValChange(): void {
```
MBuchalik, line 59: In the HTML, you write `(ionChange)="onValChange()"` pretty often. I think it would be better to do:
``` TS
ngOnInit(): void {
  this.basicInfoObj.valueChanges.subscribe(() => {
      // Your logic here
  });
}

```
in your TS file instead. (So: Subscribe to the FormGroup's valueChanges and then react to the changes here.)

frontend/src/app/forms/steps/basic-info/basic-info.component.ts
```diff
@@ -0,0 +1,71 @@
...
+    if (
+      this.basicInfoObject.controls.salutation.value.length &&
+      this.basicInfoObject.controls.firstName.value.length &&
+      this.basicInfoObject.controls.lastName.value.length
```
MBuchalik, line 63: In the future, we might think about using Validators. But let's ignore that for now. (Validators basically allow you to, for instance, mark fields as required. And then, you simply ask your form element "Are you valid?" or "Are all of your child controls valid?"

frontend/src/app/home/home.page.html
```diff
@@ -20,7 +20,9 @@ <h3>We make the difference</h3>
     </div>
 
     <div class="button-container">
-      <button class="fancy-button">Apply now</button>
+      <button class="fancy-button" (click)="onStartApplication()">
```
MBuchalik, line 23: You might want to rename it to "startApplication" or "startJobApplication".

frontend/src/app/home/home.page.ts
```diff
@@ -1,8 +1,15 @@
...
 @Component({
   selector: 'app-home',
   templateUrl: 'home.page.html',
   styleUrls: ['home.page.scss'],
 })
-export class HomePage {}
+export class HomePage {
+  constructor(private router: Router) {}
+
+  onStartApplication(): void {
```
MBuchalik, line 12: As in the comment on the html file: Please rename it to "startApplication" or "startJobApplication" or "applyNow" or whatever.

frontend/src/app/interfaces/basic-info.ts
```diff
@@ -0,0 +1,15 @@
+/*
```
MBuchalik, line 1: The name of the folder should be "model". And the file should also end on ".model.ts".

I suggest: You should move the file to app/forms/model/basic-info.model.ts or to app/forms/steps/basic-info/model/basic-info.model.ts.

MBuchalik (commented):
Good stuff! I really like the hard work you obviously put into this! 👍 🎉 

I added a few comments. There are especially a few general things regarding the commenting style and css class naming that can still be improved.

In some places, you made your life much harder that it has to be - for instance when using the ion-header. No grid or similar is needed!


But in general, a really good progress!

(I believe we can still improve the UI design in some places to better align with my mockups. But let's ignore that for now, since the design is pretty good already.)

frontend/src/app/forms/forms.page.ts
MBuchalik: Oh, and if this `@Usage` style is nothing you took from a TS guideline, then please remove these "annotations". They are pretty uncommon for TS projects (except for stuff like `@param`).

frontend/src/app/forms/forms-routing.module.ts
Bharathravi1995: I feel these comments should be there to give a fair overview of what's the component used for, info regarding Author so that someone can contact if they have doubts, Also Modifiers name to just to see who made the changes without seeing file history. With my previous experience, I witnessed it was useful and hence added it.

frontend/src/app/forms/forms.page.scss
Bharathravi1995: Ya these are the temporary changes I did. It has to be structured well. I will structure it in the coming weeks. Also whatever CSS we write it has to be responsive for different screen size. I am currently checking the possibilities to make it responsive. I feel ionic should make responsive if we handle CSS properly.

frontend/src/app/forms/forms.page.html
Bharathravi1995: Yes it was a bit messy. And does not suit all device. The main purpose having row column is to form a grid with three columns so that the elements can be positioned well. Will check for the other possibilities too.

frontend/src/app/forms/forms-routing.module.ts
MBuchalik: We have collective code ownership. So it makes no sense to say "x is the author". Also, we are using git, so you can see who did what anyways. Also: If 10 people edit the file - should they all be listed here?

If you really want to use the Modifiers, then at least write them as lowercase. (As, for instance, done in the Angular core https://github.com/angular/angular/blob/master/packages/common/src/common.ts )
But again - you usually just want to use plain comments and some built-in ones (like `@param`) that have Intellisense support.

frontend/src/app/forms/forms.page.ts
Bharathravi1995: If you did not reset it, after an user submitted the application and next user clicks on Apply now, the fields will be pre filled with previous values.

frontend/src/app/forms/forms.page.ts
Bharathravi1995: Ya I forgot it. May be we need to add a lint rule for this as well.

frontend/src/app/forms/side-menu/side-menu.component.scss
Bharathravi1995: Ya it is temporary change. Once a definite way to handle CSS , this will be changed.

frontend/src/app/forms/steps/basic-info/basic-info.component.ts
Bharathravi1995: Will remove it. Is it possible to have lints for this as well? No empty blocks something like that.If not we can write a custom lint.

frontend/src/app/forms/steps/basic-info/basic-info.component.ts
Bharathravi1995: Ya I am learning to code efficiently with Observers. I am not yet clear with it. In the next week (ionChange) will be converted as observer.

frontend/src/app/forms/steps/basic-info/basic-info.component.ts
Bharathravi1995: Ya we have to think of that. Also will be better if there is a getter setter method that can get and set value. It is pretty hard to access with object.controls.

frontend/src/app/forms/forms.page.scss
MBuchalik: I guess some Media Queries (https://developer.mozilla.org/de/docs/Web/CSS/Media_Queries/Using_media_queries) could do the trick. We can/should refactor these things partially into the global css file.

frontend/src/app/forms/forms.page.html
MBuchalik: I believe my suggestion (using the "slot" attribute) should do the trick. The Ionic docs should provide some more examples.

frontend/src/app/home/home.page.html
Bharathravi1995: This name still conveys the intended meaning. We have to discuss about naming conventions for properties, methods. We will first discuss and hear from every one. It is important to discuss ASAP before the code base becomes large. After discussion, If anythings needs to be changed I will change.

frontend/src/app/forms/steps/basic-info/basic-info.component.ts
MBuchalik: Usually, you don't need to manually set the values. But if you need to (maybe because you stored some stuff in localstorge), then setValue or patchValue on the FormGroup is your friend :)

frontend/src/app/home/home.page.ts
Bharathravi1995: We have to discuss about naming conventions for properties, methods. We will first discuss and hear from every one. It is important to discuss ASAP before the code base becomes large. After discussion, If anythings needs to be changed I will change.

frontend/src/app/forms/forms.page.ts
Bharathravi1995: I have used these kind of custom comments in my previous projects in JS. I will remove it if it is not TS way of doing it.

frontend/src/app/home/home.page.html
MBuchalik: "on" implies that something is already happening. But in your case, the application process has not started. It is the method that starts it.

This is even mentioned in the Angular Style Guide: https://angular.io/guide/styleguide#dont-prefix-output-properties

(I suggest reading the Style Guide in general. Has some valuable content.)

MBuchalik: So in most cases, I'd suggest sticking to the Style Guide or any other official documentation on the Angular website.

frontend/src/app/home/home.page.ts
MBuchalik: Same as in the other comment. (Summary: https://angular.io/guide/styleguide#dont-prefix-output-properties )

frontend/src/app/forms/forms.page.html
Bharathravi1995: We will discuss about naming conventions and change it in future if necessary.

frontend/src/app/interfaces/basic-info.ts
Bharathravi1995: Ya I will change it by tomorrow.

frontend/src/app/forms/forms.page.ts
MBuchalik: There are a few ones that are commonly used (like, as mentioned already, the `@param` annotation - you can find some others, for instance, in the Angular source code). But other than that, I would suggest using the comments starting with two asterisks (`/**`) and not using any custom annotations.

Bharathwaj-Ravi-1995:
Most of the discussions are related to naming conventions, coding style and some thing related to handling ionic property. These discussions are not stopping factor for acceptance criteria of the user story. So, I am proceeding with merging PR now for the time being to release for the current sprint. And these discussions will be resolved within next sprint.

PR #24, opened 2020-05-19
Docker usage iinstruction is updated in readme file

vishwasanavatti:
Docker repository is created for the common use of postgresql by all developers.


vishwasanavatti:
code is not proper

PR #32, opened 2020-05-22
Confirmation Page 

vishwasanavatti:


Bharathravi1995:
Looks good to me. Except for the class .secondary-color which will be changed later to color attribute. Also color  needs to be changed based on the mockup. I will do those changes. @vishwasanavatti  Once Submit page changes merged, Need to handle router transition to this view.

PR #33, opened 2020-05-22
Frontend submit

Nasser-Eddin-Nasser:


frontend/src/app/forms/forms.page.html
```diff
@@ -41,15 +41,22 @@
     >
     </form-basic-info>
   </ng-container>
+  <ng-container [ngSwitch]="currentMenu.selector">
+    <app-submit-page
```
Bharathravi1995: Will be better if all the form step component selector follow a common structure like form-<comp_purpose>

frontend/src/app/forms/forms.page.ts
```diff
@@ -108,11 +117,12 @@ export class FormsPage implements OnInit, OnDestroy {
   /*
   @Usage In this method navigation to next step is handled.
   */
-  navigateToNextForm(): void {
+  navigateToNextFormOrSubmit(): void {
     if (this.currentMenu.id !== this.totalSteps) {
       const nextMenuIndex = this.sideMenuList.indexOf(this.currentMenu) + 1;
       this.onFormStepsNavigation(this.sideMenuList[nextMenuIndex]);
     } else {
+      console.log('submit');
```
Bharathravi1995: Remove the console.log()

frontend/src/app/forms/forms.page.ts
```diff
@@ -64,6 +71,8 @@ export class FormsPage implements OnInit, OnDestroy {
   */
   private subscriptions: Subscription[] = [];
 
+  hideButton = false;
```
Bharathravi1995: Remove this property. Haven't used anywhere.

frontend/src/app/forms/forms.page.ts
```diff
@@ -48,7 +55,7 @@ export class FormsPage implements OnInit, OnDestroy {
     isActive: boolean;
     selector: string;
   };
-
+  buttonText: string;
```
Bharathravi1995: This property was not used any where now. I suggest to use this property to show continue or submit instead of calling methods in template.

frontend/src/app/forms/forms.page.ts
```diff
@@ -172,4 +182,8 @@ export class FormsPage implements OnInit, OnDestroy {
       .length;
     this.progressPercentage = completedStep / this.totalSteps;
   }
+
+  getButtonText(): string {
```
Bharathravi1995: Don't bind  method to template unless and until you cannot achieve it with property binding. Just put a console.log() inside that method and check how many times it is called . Whenever you type or change menu in any step, this method will be called because of change detection mechanism in angular.

frontend/src/app/forms/steps/submit-page/submit-page.component.ts
```diff
@@ -0,0 +1,15 @@
...
+
+@Component({
+  selector: 'app-submit-page',
+  templateUrl: './submit-page.component.html',
+  styleUrls: ['./submit-page.component.scss'],
+})
+export class SubmitPageComponent {
+  constructor(private router: Router) {}
+
+  onStartApplication(): void {
```
Bharathravi1995: What's the need for action binding in this component? Just remove the method. As of now there is no interaction in this page. It is just a static content.

Bharathravi1995:
@Nasser-Eddin-Nasser I have pushed some changes. Can you fetch the new changes and modify the eventEmitter handler name changes in html? Also comment the code. In the sideMenuList modify the id:3 instead of 2 as the contact information page will be 2nd one. I am planning to change this approach  in the near future.

frontend/src/app/forms/forms.page.ts
Nasser-Eddin-Nasser: done

frontend/src/app/forms/forms.page.html
Nasser-Eddin-Nasser: done

PR #35, opened 2020-05-25
Implement first Applicant Rating page

denoribi:
Implemented first Applicant Rating page. Changed some Global styles. Side menu content is currently hard coded. Typesafe form changed for accepting numbers.  

Did pair programming with @MBuchalik


frontend/src/app/rating/hr-side-menu/hr-side-menu.component.html
```diff
@@ -0,0 +1,32 @@
+<ion-menu side="end" contentId="hr-side-menu">
+  <ion-content>
```
Bharathravi1995, line 2: I suggest to reuse side-menu component may be we need some customisation to adopt changes. But the actions seems to be really the same.

frontend/src/app/rating/rating.page.ts
```diff
@@ -0,0 +1,29 @@
...
+  selector: 'app-rating',
+  templateUrl: './rating.page.html',
+  styleUrls: ['./rating.page.scss'],
+})
+export class RatingPage {
+  constructor(private navController: NavController) {}
+
+  ratingForm = new FormGroup<RatingForm>({
+    applicantScore: new FormGroup<ApplicantScore>({
+      rhetoric: new FormControl(undefined, Validators.required),
```
Bharathravi1995, line 28: I have a doubt. Why we need to define the type as undefined? Can't we use number? since the rhetoric type is defined as number in interface.

frontend/src/app/rating/steps/applicant-score/applicant-score.component.html
```diff
@@ -0,0 +1,65 @@
+<h1>Applicant Rating</h1>
+
+<ion-list lines="none">
+  <ion-list-header>
+    Rhetoric:
+    {{ ratingForm.controls.applicantScore.controls.rhetoric.value }}
+  </ion-list-header>
+  <ion-item class="rating">
```
Bharathravi1995, line 8: May be we can implement this as an array of objects. And use ngFor to iterate in the template. But need to make sure two way binding happens when value changes. The current way of doing will become cumbersome if we add more list.

frontend/src/app/rating/rating.module.ts
```diff
@@ -0,0 +1,21 @@
+import { CommonModule } from '@angular/common';
```
Bharathravi1995, line 6: Can you add comments for all the files? You can check my previous PR for commenting procedure.

Bharathravi1995 (commented):
Overall the PR is clean. Good job @denoribi .

frontend/src/app/rating/hr-side-menu/hr-side-menu.component.html
MBuchalik: The problem is that the code related to HR is completely different compared to the one used for the "main" form. For instance, this side menu already uses Validators etc. So from my point of view, it is currently not trivial to make the side menus compatible with each other.

My suggestion: Leave it as it is and when the main form has been migrated to Validators etc, then we can think about merging it into one Component.

frontend/src/app/rating/rating.page.ts
MBuchalik: Usually, if you don't want to pre-fill a form element with a value, then you simply leave it blank:
``` ts
rhetoric: new FormControl(),
```

But: Validators are passed as second parameter. So, if you want to use a Validator, you have to write:

``` ts
rhetoric: new FormControl(SOMETHING, Validators.required),
```

If you pass a number as first parameter, then the form element will be pre-filled with this value. And the Validator will automtically mark this Form Control as valid. This is pretty bad UX, since the user is now not forced to change the input value.

So, if you don't want to pre-fill the form element with a value, but still want to use a FormControl, then you need to do it like it was done here.

frontend/src/app/rating/steps/applicant-score/applicant-score.component.html
MBuchalik: I think this is something we could refactor later... We need to think about how to implement various other things (like routing and stuff like that), so we could refactor it then I guess.

(For heterogenous data (like mixed text fields, sliders, select menus etc.), refactoring into an array of objects won't work well anyways... So most of our components will look like this anyways.)

frontend/src/app/rating/rating.module.ts
denoribi: i will add some comments :)

PR #36, opened 2020-05-25
database access for HR-Monitor

vishwasanavatti:
Established a connection to show the stored data in the HR-Monitor


MBuchalik:
I just had a brief look at the code (someone with more backend experience should review this), and I have a question:

Did I understand it correctly that you are connecting directly to the database with the HR application? If so: The HR application will be installed on the computers of the HR people, right? And the database will run on some server in the cloud. So you basically need to expose the database over the internet, so that the HR applications can access it? Sounds pretty dangerous to me... Wouldn't it be better to build a JSON API (or similar)?

Nasser-Eddin-Nasser (commented):
I pulled, ran the code on my pc, it works correctly :)

MBuchalik:
@vishwasanavatti Please have a look at my previous comment. I got the feeling this is the wrong approach here.

vishwasanavatti:
@MBuchalik You might have understood it differently but JSON API is really not necessary. We can discuss tomorrow during the meeting.

MBuchalik:
@vishwasanavatti Alright 👍 I just wanted to make sure you aren't planning to expose a database directly to the internet :D

PR #37, opened 2020-05-26
Basic CRUD operations handled for using ionic storage

Bharathravi1995:
**Done:**

- A basic CRUD operations methods are handled to use ionic storage easily.
- Interfaces moved as models and each steps are defined under overall info model.

**To do:**

- Error handling needs to be done
- Async await support in functions
- Backup mechanism to store in sqlite when indexed db / local storage is cleared by OS.
- Storing images and files. As of now storing images as blob and base 64 formats are studied. But both has its own consequences. As indexed db supports ArrayBuffer formats , should try to use that.

**Issues faced:**
While choosing Iphone, when the first data is stored in indexed db, on page refresh the storage shifts to local storage. But doesn't happen with android. Because of this shift, data stored in the indexed db can't be retrieved as local forage fetched data only from the current storage system.

**Further clarifications needed:**
Quota allocation needs to be understood well, as it is dynamic in nature. From the guides 50% of free disk space is allocated for browser. The fraction for each storage system is unclear. Also most of the data available for web application nothing much available for mobile apps and mobile OS.


frontend/package.json
```diff
@@ -30,6 +30,7 @@
     "@ionic-native/splash-screen": "^5.24.0",
     "@ionic-native/status-bar": "^5.24.0",
     "@ionic/angular": "^5.1.0",
+    "@ionic/storage": "^2.2.0",
```
MBuchalik, line 33: Don't forget to add it to the Bill of Materials then :)

frontend/src/app/app.module.ts
```diff
@@ -4,14 +4,23 @@ import { RouteReuseStrategy } from '@angular/router';
...
 @NgModule({
   declarations: [AppComponent],
   entryComponents: [],
-  imports: [BrowserModule, IonicModule.forRoot(), AppRoutingModule],
+  imports: [
+    BrowserModule,
+    IonicModule.forRoot(),
+    IonicStorageModule.forRoot({
+      name: 'digitact',
+      driverOrder: ['indexeddb', 'localstorage'],
```
MBuchalik, line 20: Have you found a way to detect which driver is currently being used? At least in browsers, localstorage has a quota of a few megabytes. If the storage system switches to it, then we will be screwed (since, after storing a few images, the storage will be full.) I guess we should in this case show an error message or so saying that someting is super wrong :D

frontend/src/app/forms/forms.page.ts
```diff
@@ -138,6 +137,14 @@ export class FormsPage implements OnInit, OnDestroy {
...
       this.navigateToStep(this.sideMenuList[nextMenuIndex]);
     } else {
+      const key = (
+        this.overallInfo.value.basicInfo.firstName +
+        '-' +
+        this.overallInfo.value.basicInfo.lastName +
+        '-' +
+        Math.floor(Math.random() * 1000000).toString()
+      ).replace(/\s+/g, '_');
+      this.storage.addItem(key, this.overallInfo.value).then(() => {});
```
MBuchalik, line 147: Fine for now. But for the future, I guess we should improve this (as you already mentioned). I think it would be great just to use an ID (auto-increment fashion) for the items.

Also, it would be really cool if we could use more than one table. Not sure if this is easily possible in Ionic Storage. But it would help a lot if there was one table for the Job Applications, one for some general settings, ...

frontend/src/app/model/forms-data.model.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * @description
+ *    This model holds the all the interfaces related to forms data.
+ */
```
MBuchalik, line 4: Important: When you write a comment like this, then it will always show up when someone hovers the "FormsData" interface. But I guess you want this comment only to be visible within this file, right? In this case, it would make more sense not to use a `/**` comment, but rather a normal `/*` comment. 

So, in general: If you add a comment to a specific element (e.g. a class, interface, method, class field, ...), then use the `/**` comments. But if you write a "general" comment that is not tied to a specific element, then use the `/*` comment. 

You can think of it like this: "Do I want the comment to show up when someone hovers the method name, class field, whatever?" Then, use the `/**` comment. But if you don't want it to show up when someone hovers the name of the next element (e.g. because this is not directly referring to the item, but is rather an overall explanation), then use the `/*` one. 

I hope that is somewhat understandable? Sorry if I haven't explained it properly in the previous discussions.

You don't need to change it now - just keep it in mind for the future :)

frontend/src/app/services/storage-handler.service.ts
```diff
@@ -0,0 +1,108 @@
...
+          return data;
+        });
+      });
+    });
+  }
+}
+
+type DeepPartial<T> = {
+  [key in keyof T]?: DeepPartial<T[key]>;
+};
```
MBuchalik, line 108: There is also a DeepPartial in my custom Forms file. Maybe, we should put it in a separate model in the future. But you don't have to do this now :)

frontend/src/app/services/storage-handler.service.ts
```diff
@@ -0,0 +1,108 @@
+/**
```
MBuchalik, line 1: I am not sure if we should name the folder "services" or just "service". (Similar to "model" vs. "models") What do you think?

I just looked it up on the internet since I always used the singular form. Originally, I thought that naming the folders "model" and "service" would be part of the Angular Style Guide. But I actually didn't find it here. So I simply don't know where I got the ideas for naming the folders from. There are a few projects that do it in a relatively similar fashion (e.g. https://github.com/gothinkster/angular-realworld-example-app)

Sorry for the confusion :/

MBuchalik (commented):
Looking really good! 👍 🎉 

I added a few review comments, but if you have read them, you can simply merge this PR if you want to 👍 

It would be nice if you could create an Issue where we list the things we still need to do. Stuff that comes to my mind (partially as you have already mentioned):
- Use SQLite so that we are sure data won't be lost
- Find a way to use multiple tables
- Implement error handling
- Store the data once someone clicks on the "submit" button (I have only had a brief look at it, but currently it is stored before, am I right?)

frontend/src/app/forms/forms.page.ts
MBuchalik: Btw: I know that auto-increment is not possible in Ionic Storage. But we could maybe just use a separate storage item for that :) (so: an item that holds the current "auto increment" number)

Bharathravi1995: Ya It would be good if we have more db. It can be structured really well, If we achieve it. There is an open issue related to ionic storage in it https://localforage.github.io/localForage/#multiple-instances-createinstance . I have started checking to achieve this with localforage.

frontend/src/app/services/storage-handler.service.ts
Bharathravi1995: Ya thought the same. I am still not yet clear with DeepPartial. Some time we should catch up on a meeting through zoom.

frontend/src/app/services/storage-handler.service.ts
```diff
@@ -0,0 +1,108 @@
...
+      });
+    });
+  }
+
+  /**
+   * Used to get all items from the storage
+   * @returns Promise of array of forms data object
+   */
+  getAllItems(): Promise<FormsData[]> {
+    const items: Array<FormsData> = [];
```
MBuchalik, line 48: Maybe, just use `FormsData[]` instead of `Array<FormsData>`. I don't think there is a technical difference, but I got the feeling I see the first version much more often in other projects

frontend/src/app/forms/forms.page.ts
MBuchalik: Wondering if this works: https://stackoverflow.com/questions/55470808/ionic-3-multiple-tables-in-indexeddb

Then we could have multiple Services (one for each table)

frontend/src/app/services/storage-handler.service.ts
Bharathravi1995: I uses plural form if there gonna be more services under it. Currently it is not there, but in future we may have more services. I thought it is just similar to assets environments folder.

frontend/src/app/services/storage-handler.service.ts
MBuchalik: Sure 👍 Or we can also discuss it in Slack.

frontend/src/app/app.module.ts
Bharathravi1995: https://ionicframework.com/docs/angular/storage#driver using this we can know the current driver in usage. But this issue exists only ios as far as I tested. We need to somehow try to access content in all the storage sysytem. It can be accessed using indexeddb and localstorage API dorectly just like we do in normal web apps. But this case needs to handled really carefully. Will check.

frontend/package.json
Bharathravi1995: Oh yeah I forgot. Thanks for pointing out.

frontend/src/app/model/forms-data.model.ts
Bharathravi1995: > Looking really good! 👍 🎉
> 
> I added a few review comments, but if you have read them, you can simply merge this PR if you want to 👍
> 
> It would be nice if you could create an Issue where we list the things we still need to do. Stuff that comes to my mind (partially as you have already mentioned):
> 
> * Use SQLite so that we are sure data won't be lost
> * Find a way to use multiple tables
> * Implement error handling
> * Store the data once someone clicks on the "submit" button (I have only had a brief look at it, but currently it is stored before, am I right?)

I will file issues for better tracking.

frontend/src/app/app.module.ts
MBuchalik: Maybe, the issue you saw was just a temporary glitch? I am just wondering how such a popular library can have such a bug in it :D 

Anyways: I guess as soon as we have integrated SQLite, the issue should be gone

frontend/src/app/services/storage-handler.service.ts
Bharathravi1995: It does not make any difference. I simply liked the generic approach of declaring array in typescript. Hence followed it.

frontend/src/app/app.module.ts
Bharathravi1995: May be. SQlite should be a saviour as of now.

frontend/src/app/forms/forms.page.ts
Bharathravi1995: Will check

MBuchalik:
@Bharathravi1995 From my point of view, you can merge it. I wasn't sure if you want to do add anything here, so I didn't merge it.

Btw: You are doing a lot of work/reasearch here. I believe this takes a lot of time. --> Thumbs up 👍 👍 👍 Maybe, we could split our work so that you can concentrate more on the Storage issues and I concentrate more on the UI/Forms? I just want to make sure you don't get burned out :D Storage can become pretty complex...

denoribi (commented):
Looks good for my understanding ;) Pretty clean PR.

Bharathravi1995:
> @Bharathravi1995 From my point of view, you can merge it. I wasn't sure if you want to do add anything here, so I didn't merge it.
> 
> Btw: You are doing a lot of work/reasearch here. I believe this takes a lot of time. --> Thumbs up 👍 👍 👍 Maybe, we could split our work so that you can concentrate more on the Storage issues and I concentrate more on the UI/Forms? I just want to make sure you don't get burned out :D Storage can become pretty complex...

Yes we can split it and work. Now I gonna merge this PR.

PR #42, opened 2020-05-28
Use Validators

MBuchalik:
I updated the main form to now use (and rely on) Validators. This involved implementing quite a lot of (structural) changes, so I also refactored some of the related code.

(I also slightly enhanced two of the Steps. I did this during the general refactoring, so I wasn't really able to create a separate commit for that.)

The diff is pretty hard to read. I'd suggest to simply pull and check the changes locally.

@Bharathravi1995 I guess you will be the one to review this. I know that there is still a lot of room for improvement. However, most of the developers cannot really start their work until this PR has been merged. (That's why I implemented this so quickly.) So: I'd suggest to merge this PR if there are no critical/fundamental problems and to ignore stuff like maybe missing comments or similar for now. :)

### DIFF
--- a/frontend/src/app/forms/forms.page.html
+++ b/frontend/src/app/forms/forms.page.html
@@ -6,9 +6,15 @@
       </ion-button>
     </ion-buttons>
 
-    <ion-title class="ion-text-center"
-      >Step {{currentMenu.id}}/ {{totalSteps}}</ion-title
-    >
+    <ion-title class="ion-text-center">
+      <ng-container *ngIf="currentStep !== APPLICATION_STEP.Submit">
+        Step {{currentStepIndex + 1}} / {{APPLICATION_STEP_ARR.length - 1}}
+      </ng-container>
+
+      <ng-container *ngIf="currentStep === APPLICATION_STEP.Submit">
+        Submit
+      </ng-container>
+    </ion-title>
 
     <ion-buttons slot="end">
       <ion-menu-button autoHide="true" menuClose id="main">
@@ -20,38 +26,50 @@
 </ion-header>
 
 <app-side-menu
-  [sideMenuList]="sideMenuList"
-  (chosenStep)="navigateToStep($event)"
->
-</app-side-menu>
+  [formsData]="formsData"
+  [currentStep]="currentStep"
+></app-side-menu>
 
 <ion-content>
   <div class="main-container with-bottom-buttons">
     <div class="content">
-      <ng-container [ngSwitch]="currentMenu.selector">
+      <ng-container [ngSwitch]="currentStep">
         <form-basic-info
-          (updatedMenuCompletionStatus)="updateProgessStatus()"
-          [menuObject]="currentMenu"
-          [basicInfoObject]="overallInfo.controls.basicInfo"
-          *ngSwitchCase="'form-basic-info'"
+          [formsData]="formsData"
+          *ngSwitchCase="APPLICATION_STEP.BasicInformation"
         >
         </form-basic-info>
         <form-contact-info
-          (updatedMenuCompletionStatus)="updateProgessStatus()"
-          [menuObject]="currentMenu"
-          [contactInfoObject]="overallInfo.controls.contactInfo"
-          *ngSwitchCase="'form-contact-info'"
+          [formsData]="formsData"
+          *ngSwitchCase="APPLICATION_STEP.ContactInformation"
         >
         </form-contact-info>
 
-        <form-submit-page *ngSwitchCase="'form-submit-page'">
+        <form-submit-page
+          [formsData]="formsData"
+          *ngSwitchCase="APPLICATION_STEP.Submit"
+        >
         </form-submit-page>
       </ng-container>
     </div>
 
     <div class="buttons-container">
-      <button class="fancy-button" (click)="navigateToNextStep()">
-        {{buttonText}}
+      <ng-container *ngIf="currentStep === APPLICATION_STEP.Submit">
+        <button
+          class="fancy-button"
+          [disabled]="!formsData.valid"
+          (click)="submit()"
+        >
+          Submit
+        </button>
+      </ng-container>
+
+      <button
+        class="fancy-button"
+        (click)="navigateToNextStep()"
+        *ngIf="currentStep !== APPLICATION_STEP.Submit"
+      >
+        Continue
       </button>
     </div>
   </div>
--- a/frontend/src/app/forms/forms.page.ts
+++ b/frontend/src/app/forms/forms.page.ts
@@ -163,7 +163,7 @@ export class FormsPage implements OnInit, OnDestroy {
   }
 
   /**
-   * Update the "step" query paremeter.
+   * Update the "step" query parameter.
    * You can use this to navigate between the different form steps!
    */
   navigateToStep(step: ApplicationStep): void {
--- a/frontend/src/app/forms/forms.page.ts
+++ b/frontend/src/app/forms/forms.page.ts
@@ -1,18 +1,20 @@
-/**
- * @description
- *   This page handles the basic operation of tracking progress, navigation, close menu, continue button,
- *   event handlers for child to parent communication and parent to child data down.
- */
+/*
+  @description
+    This page handles the basic operation of tracking progress, navigation, close menu, continue button,
+    event handlers for child to parent communication and parent to child data down.
+*/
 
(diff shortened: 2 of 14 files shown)


frontend/src/app/forms/forms.page.html
```diff
@@ -6,9 +6,15 @@
       </ion-button>
     </ion-buttons>
 
-    <ion-title class="ion-text-center"
-      >Step {{currentMenu.id}}/ {{totalSteps}}</ion-title
-    >
+    <ion-title class="ion-text-center">
+      <ng-container *ngIf="currentStep !== APPLICATION_STEP.Submit">
```
Bharathravi1995, line 10: According to me template conditionals should be avoided whenever it is easily achievable with TS. We can have property titleText and update it according to the step , as anyways we will be listening to the query change.

frontend/src/app/forms/forms.page.html
```diff
@@ -20,38 +26,50 @@
...
+          [formsData]="formsData"
+          *ngSwitchCase="APPLICATION_STEP.BasicInformation"
         >
         </form-basic-info>
         <form-contact-info
-          (updatedMenuCompletionStatus)="updateProgessStatus()"
-          [menuObject]="currentMenu"
-          [contactInfoObject]="overallInfo.controls.contactInfo"
-          *ngSwitchCase="'form-contact-info'"
+          [formsData]="formsData"
```
Bharathravi1995, line 43: Why the whole formsData should be sent to component. Shouldn't be enough if only required step info is passed?

frontend/src/app/forms/forms.page.html
```diff
@@ -20,38 +26,50 @@
...
       </ng-container>
     </div>
 
     <div class="buttons-container">
-      <button class="fancy-button" (click)="navigateToNextStep()">
-        {{buttonText}}
+      <ng-container *ngIf="currentStep === APPLICATION_STEP.Submit">
+        <button
+          class="fancy-button"
+          [disabled]="!formsData.valid"
```
Bharathravi1995, line 60: Good use of validators.

frontend/src/app/forms/forms.page.ts
```diff
@@ -21,196 +23,208 @@ import { FormControl, FormGroup } from './../common/forms/forms';
...
+  formsData = new FormGroup<FormsData>({
+    basicInfo: new FormGroup<BasicInfo>({
+      firstName: new FormControl('', Validators.required),
+      lastName: new FormControl('', Validators.required),
+      salutation: new FormControl(undefined, Validators.required),
+    }),
+    contactInfo: new FormGroup<ContactInfo>({
+      phoneNumber: new FormControl('', Validators.required),
+      eMail: new FormControl('', Validators.required),
+      linkedIn: new FormControl(''),
```
Bharathravi1995, line 53: Optional fields handling became easy with validators. Like this approach.

frontend/src/app/forms/side-menu/side-menu.component.html
```diff
@@ -9,32 +9,77 @@ <h1>Progress</h1>
         </div>
       </div>
 
-      <ion-list lines="none" *ngFor="let menu of sideMenuList">
-        <ion-menu-toggle auto-hide="true">
+      <ion-list lines="none">
+        <ng-container
```
Bharathravi1995, line 13: Should be better if it can be done in ngFor. Whenever a new step is added we have to add ng-container here as well which can be avoided if we go Array of objcets approach but we need to bind form control valid property to isValid before opening this component. We will think and iterate it as we move forward.

frontend/src/app/forms/side-menu/side-menu.component.html
```diff
@@ -9,32 +9,77 @@ <h1>Progress</h1>
...
+          [ngTemplateOutletContext]="{
+            step: APPLICATION_STEP.ContactInformation,
+            displayName: 'Contact Information',
+            isValid: formsData.controls.contactInfo.valid,
+            isActive: currentStep === APPLICATION_STEP.ContactInformation
+          }"
+        ></ng-container>
+
+        <ion-menu-toggle>
           <ion-item
```
Bharathravi1995, line 34: Could have been added in ng-container as only the icon name varies which can be handled in the ternary check.

frontend/src/app/forms/side-menu/side-menu.component.ts
```diff
@@ -1,43 +1,61 @@
...
    */
-  changeStep(menu: {
-    id: number;
-    displayName: string;
-    isCompleted: boolean;
-    isActive: boolean;
-    selector: string;
-  }): void {
-    this.chosenStep.emit(menu);
+  navigateToStep(step: ApplicationStep): void {
```
Bharathravi1995, line 53: I prefer not to repeat the code and handle step transition in one place by emitting event to parent method.

Bharathravi1995 (commented):
@MBuchalik Good job!  Overall Validators helped a lot by avoiding boilerplates. But in some places I felt using array of objects could have made life simpler. I added some suggestions to my visibility. You can merge this PR if u don't want to add anything Or I can also merge it.

MBuchalik:
Thanks for your review :) I will merge it now, since I saw no major concerns in your comments. I am currently in a hurry, so I cannot reply to your comments now. I will do so later or maybe tomorrow. :)

frontend/src/app/forms/forms.page.html
MBuchalik: (This particular example is somewhat "special", so I will comment on the general approach and not this very example.)

I believe that it is less error-prone if you have only one single property in your TS file for a specific "task". For example, in our case, we have the "currentStep". In our TS file, we can "blindly" update this property in any way we want. When it gets updated, then our template will automatically "follow" and display the correct header text. But if we had two properties ("currentStep" and "titleText"), then we would need to be very careful when updating the "currentStep", since we should make sure not to forget to update "titleText". Of course, you can solve this by using some separate methods like "setCurrentStep" or so. But usually, you just want to directly say `this.currentStep = foo`. I see "currentStep" as some sort of "data" or "data model", while I see the title text as a "presentational" thing. The TS file should most of the times hold only "data" items and not really care about the "presentation", while the template does the opposite.

Of course, dividing it like this is not always possible. But in general, I believe that using this approach is less error-prone.

But honestly - it doesn't really matter if we use this approach or the one you suggested. Both are fine. :) I originally wanted to use a TS setter for that, but I guess calling a TS getter in the template is not really ideal (performance wise).

frontend/src/app/forms/forms.page.html
MBuchalik: Please have a look at contact-info.component.html. Here, it says "Nice to meet you <name>" (if a name has already been entered). I was able to easily achieve this since I had access to the whole formsData object. Another example: Please have a look at submit-page.component.html Depending on whether the form is valid or not, a different text is displayed.

So, by passing the whole formsData object to the individual components, you can easily create all kinds of "advanced" interactions utilizing form data entered in different steps.

I don't see any disadvantage from passing the whole object to the components (you just have to write "formsData.controls" pretty often :D ). But the advantage is that you can easily implement all kinds of interactions that wouldn't be possible if we had only one "slice" of this object.

I'd suggest to pass this object to all components and never to use the individual step info data. This allows us to easily add fancy stuff easily without having to refactor the whole component just because I want to display the user's name ;)

frontend/src/app/forms/side-menu/side-menu.component.html
MBuchalik: We can definitely improve this solution. But: If you create an object in ngOnInit that references an input property, then you can introduce all kinds of funny bugs. Imagine the input property changes - then this won't be reflected in the object we created in ngOnInit. So, the current solution is at least more "robust" and less error-prone.

frontend/src/app/forms/side-menu/side-menu.component.html
MBuchalik: The ng-template also has properties like "isValid" that aren't really useful here, so I felt that adding a separate item allows us to later change/style it later much easier. But we can also move it into the ng-template - I don't really have a strong opinion about that.

frontend/src/app/forms/side-menu/side-menu.component.ts
MBuchalik: I don't really have a strong opinion about that. To me, it feels like the purpose of the router is, among others, to allow changing the route in various places. But we can also emit an event to the parent. :)

PR #43, opened 2020-05-30
Add a feature documentation and convert readme files to pdf

MBuchalik:
I added a feature documentation. Also, I converted some md files to pdf.

I personally don't really see any benefit from using pdf files, but we were requested to provide them. Maybe, we can simply delete them afterwards. :D (Only using MD files has the advantage that you can easily edit them without having to convert them to pdf all the time. Also, you can view md files on GitHub easily, so I really only see disadvantages here.)


denoribi:
Looks good for me, but I'm not sure, if we need to change the pdf and add our new pages from this week too. So i will merge this now and if we want to add the new pages(if they are all finished), we can change it then.

PR #44, opened 2020-05-30
Db schema

Nasser-Eddin-Nasser:


Backend/README.md
```diff
@@ -86,3 +86,39 @@ Log in using:
...
+    "public_Sector": boolean
+  },
+  "positions": {
+    "consultant_Business_Consultant": boolean,
+    "iT_Consultant_Informationsmanagement": boolean,
+    "iT_Consultant_Java_JEE": boolean,
+    "iT_Consultant_Data_Science": boolean,
+    "iT_Consultant_Artificial_Intelligence": boolean,
+    "internship_Working_Student": boolean,
+    "consultant_SAP": boolean
```
MBuchalik, line 111: So for every possible Position, you need to send a boolean to the server? Wouldn't it be better if you could simply send a string array to the server? Imagine we have 100 possible positions - then you also need to provide 100 values, even if the user only wants to work in 1 position... The same goes for the columns - you need n columns for n possible values. :/ Having a separate table (and using SQL Joins) could probably make things much easier :) What do you think?

Backend/build.gradle
```diff
@@ -1,34 +1,36 @@
...
 }
 
 dependencies {
-	implementation 'org.springframework.boot:spring-boot-starter-web'
-	implementation 'org.springframework.boot:spring-boot-starter-data-jpa'
-	runtimeOnly 'org.postgresql:postgresql'
-	testImplementation('org.springframework.boot:spring-boot-starter-test') {
-		exclude group: 'org.junit.vintage', module: 'junit-vintage-engine'
-	}
+    compile 'com.google.googlejavaformat:google-java-format:1.6'
```
MBuchalik, line 16: Please remember to upate the Bill of Materials once the PR has been accepted :)

Backend/README.md
Nasser-Eddin-Nasser: It depends on the data type, at the moment its a fixed data, so having it as boolean is the best solution for the DB.

Nasser-Eddin-Nasser: I will provide also a DB schema as a table so it will make it better to understand :)

Nasser-Eddin-Nasser: DB-Schema

Bharathravi1995: > So for every possible Position, you need to send a boolean to the server? Wouldn't it be better if you could simply send a string array to the server? Imagine we have 100 possible positions - then you also need to provide 100 values, even if the user only wants to work in 1 position... The same goes for the columns - you need n columns for n possible values. :/ Having a separate table (and using SQL Joins) could probably make things much easier :) What do you think?

@Nasser-Eddin-Nasser Even I feel it is better to have array of strings rather than a field for everything. The problem here is whenever a new field is added , backend API won't be backward compatible. You may need hack to support it backward compatible which may slow down the development. Also in backend will there gonna be any specific business logic related to these fields? I don't think at this moment u need to have specific handling for Automotive field or AI or DataScience in the backend.

Nasser-Eddin-Nasser: Hey, I also agree on it but, Anshaj means the Positions and Industry are fixed values (will not be changed), that's why I did it this way so that in our  DB we will store each position/Industry as a column with boolean values (less store capacity needed, and more efficient while having DB operations). 
In case we will store it as a string array values inside the DB we will have to insert more values in our DB table in one column (that's bad).

Another way to do is to map all the DB data to Objects in our BE Server and then make the whole operations we want and then update the data (more efficient while having DB operations because it will be made in Server and more capacity needed in the Server).

The problem is we still don't know how/who well change these values (Usually it should be updated/added in the BE and then by init new client/ client refresh, the values will be sent to the client (FE)  where it will be rendered there )! 

Sure we still can change it, it's not our last sprint :)

Nasser-Eddin-Nasser: @Bharathravi1995  @MBuchalik  Sure, thank you both for reviewing the Pull req.  :)

Nasser-Eddin-Nasser: DB-Schema-2

Nasser-Eddin-Nasser: After having a meeting with @dalex135  and @vishwasanavatti  we decided to make DB schema "smarter", with a mapping function. the mapping will be done in the BE Server.  
please let me know if you have any questions or suggestions @MBuchalik  @Bharathravi1995

MBuchalik: In the "Industry" (and "Position") field of the first row, you wrote "1,2". Does that mean that you will store more than one value per field? What about normalizing it by creating a separate table (as suggested in my first comment)? Would this have some downsides? You wouldn't really need the mapping table then as well, since you could simply store arbitrary strings here.

I mean somthing like this:

Main table:
| ID | Phone Number | Email       |
|----|--------------|-------------|
| 1  | 12345678     | example@example.com |
| 2  | 987654321    | example@example.com |

"Positions" table:
| ID | Applicant ID (foreign key) | Position             |
|----|----------------------------|----------------------|
| 1  | 1                          | Some Position        |
| 2  | 1                          | Another Position     |
| 3  | 1                          | Yet another Position |
| 4  | 2                          | Some Position        |

What would be the downside of this?

Nasser-Eddin-Nasser: In case of having several values per field is the worst thing to do, but because some developers pointed out that Position & industry are not fixed (not in the requirement), so we discussed it together, and we decided to change our schema and to add a mapping layer.  
 In your example, the same position value will be stored several times on different users, in case we only have a few users (around 100) it will be fine otherwise we will have too big DB.
 What's the problem with the mapping structure?

MBuchalik: The mapping structure itself is ok. I am just concerned about the fact that you want to store multiple values in a single field. Or am I interpreting your diagram in the wrong way? (I don't know how Postgres works in detail. Maybe, this is just how you usually do it. I can only speak for other SQL databases and here, you always try to normalize your database. Maybe not to the highest normal form, but at least try not to violate 1NF.)

Btw, my example could also be updated to take your examples into account:

| ID | Applicant ID (foreign key) | Position (references the mapping table) |
|----|----------------------------|-----------------------------------------|
| 1  | 1                          | 1                                       |
| 2  | 1                          | 3                                       |
| 3  | 1                          | 9                                       |
| 4  | 2                          | 3                                       |


Anyways - I see that this Sprint is nearly over, so we should maybe just merge it and change it later...

dalex135 (commented):
Checked the changed files and the DB schema. We should have a discussion on DB schema with all the developers. I think we can merge this.

Backend/README.md
Nasser-Eddin-Nasser: This table looks very good, and we can map it directly in the DB, I'm gonna implement it in the next sprint  :)

vishwasanavatti (approved):
checked locally, works fine.

PR #45, opened 2020-05-31
Relevant industry and position

denoribi:
After @Bharathravi1995 implemented the basic skeleton of the field and designation page and added a workaround to a problem with FormArray, i did fill in the correct information from our "Jobs and Industry" document, implemented the design and changed the design of the popover. 

Because we still got a problem with FormArray, which we wanted to use, it could be necessary to rework the page after we found a fix for this problem.

### DIFF
--- a/frontend/src/app/forms/steps/field-designation-page/field-designation-page.component.html
+++ b/frontend/src/app/forms/steps/field-designation-page/field-designation-page.component.html
@@ -1,12 +1,12 @@
 <div class="main-description">
-  <h1>Field and Designation</h1>
+  <h1>Industries and Positions</h1>
   <span>
     So that we can know in which industries and positions you are intrested,
     please choose from the following.
   </span>
 </div>
 
-<ion-item class="ion-no-padding">
+<ion-item class="ion-no-padding form-item">
   <ion-label color="medium">Industry</ion-label>
   <ion-select
     multiple
@@ -17,12 +17,12 @@
       *ngFor="let field of industryArray"
       [value]="field.value"
     >
-      {{ field.displayName }}</ion-select-option
-    >
+      {{ field.displayName }}
+    </ion-select-option>
   </ion-select>
 </ion-item>
 
-<ion-item class="ion-no-padding">
+<ion-item class="ion-no-padding form-item">
   <ion-label color="medium">Position</ion-label>
   <ion-select
     multiple
@@ -33,7 +33,7 @@
       *ngFor="let designation of positionsArray"
       [value]="designation.value"
     >
-      {{ designation.displayName }}</ion-select-option
-    >
+      {{ designation.displayName }}
+    </ion-select-option>
   </ion-select>
 </ion-item>
--- a/frontend/src/app/forms/steps/field-designation-page/field-designation-page.component.html
+++ b/frontend/src/app/forms/steps/field-designation-page/field-designation-page.component.html
@@ -1,0 +1,1 @@
+<p>Add fields and designation multi slect menus</p>
--- a/frontend/src/app/forms/steps/field-designation-page/field-designation-page.component.html
+++ b/frontend/src/app/forms/steps/field-designation-page/field-designation-page.component.html
@@ -1,5 +1,5 @@
 <div class="main-description">
-  <h1>Industries and Positions</h1>
+  <h1>Field and Designation</h1>
   <span>
     So that we can know in which industries and positions you are intrested,
     please choose from the following.
@@ -11,8 +11,7 @@
   <ion-select
     multiple
     [formControl]="formsData.controls.fieldDesignationInfo.controls.field"
-    [interfaceOptions]="{ cssClass: 'custom-popover' }"
-    interface="popover"
+    [interfaceOptions]="{ cssClass: 'custom-alert' }"
   >
     <ion-select-option
       *ngFor="let field of industryArray"
@@ -28,8 +27,7 @@
   <ion-select
     multiple
     [formControl]="formsData.controls.fieldDesignationInfo.controls.designation"
-    [interfaceOptions]="{ cssClass: 'custom-popover' }"
-    interface="interface"
+    [interfaceOptions]="{ cssClass: 'custom-alert' }"
   >
     <ion-select-option
       *ngFor="let designation of positionsArray"
--- a/frontend/src/app/forms/steps/field-designation-page/field-designation-page.component.html
+++ b/frontend/src/app/forms/steps/field-designation-page/field-designation-page.component.html
@@ -1,1 +1,41 @@
-<p>Add fields and designation multi slect menus</p>
+<div class="main-description">
+  <h1>Industries and Positions</h1>
+  <span>
+    So that we can know in which industries and positions you are intrested,
+    please choose from the following.
+  </span>
+</div>
+
+<ion-item class="ion-no-padding">
+  <ion-label color="medium">Industry</ion-label>
+  <ion-select
+    multiple
+    [formControl]="formsData.controls.fieldDesignationInfo.controls.field"
+    [interfaceOptions]="{ cssClass: 'custom-popover' }"
+    interface="popover"
+  >
+    <ion-select-option
+      *ngFor="let field of industryArray"
+      [value]="field.value"
+    >
+      {{ field.displayName }}</ion-select-option
+    >
+  </ion-select>
+</ion-item>
+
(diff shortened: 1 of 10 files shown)


frontend/src/app/common/forms/forms.ts
```diff
@@ -89,7 +89,7 @@ type InitialValueOrInitialValueAndDisabled<T> =
  * For an example, see comment on the FormGroup wrapper class.
  */
 export class FormControl<
-  T extends string | number | File
+  T extends string | number | File | []
```
MBuchalik: Yeah, that is probably because the typsafe implementation currently cannot handle arrays on FormControls. This change will become irrelevant once we have found a general approach for handling arrays in FormControls.

frontend/src/app/forms/steps/field-designation-page/field-designation-page.component.html
```diff
@@ -0,0 +1,41 @@
...
+  </span>
+</div>
+
+<ion-item class="ion-no-padding">
+  <ion-label color="medium">Industry</ion-label>
+  <ion-select
+    multiple
+    [formControl]="formsData.controls.fieldDesignationInfo.controls.field"
+    [interfaceOptions]="{ cssClass: 'custom-popover' }"
+    interface="popover"
```
MBuchalik: On the browser's console, you will see the warning 
```
Select interface cannot be "popover" with a multi-value select. Using the "alert" interface instead.
```
So it would be better to remove `interface="popover"`. Also, maybe rename the css class names like "custom-popover" (since this is not a popover anymore :) )

frontend/src/app/forms/steps/field-designation-page/field-designation-page.component.html
```diff
@@ -0,0 +1,41 @@
...
+    <ion-select-option
+      *ngFor="let field of industryArray"
+      [value]="field.value"
+    >
+      {{ field.displayName }}</ion-select-option
+    >
+  </ion-select>
+</ion-item>
+
+<ion-item class="ion-no-padding">
```
MBuchalik: I ran it and saw that there is not much spacing between these elements. Maybe, slightly increase the margin between the two inputs? (Tbh, I am super picky here :D If you find time, maybe do it. But this is not super important.)

frontend/src/app/forms/steps/field-designation-page/field-designation-page.component.html
```diff
@@ -0,0 +1,41 @@
...
+  <ion-label color="medium">Position</ion-label>
+  <ion-select
+    multiple
+    [formControl]="formsData.controls.fieldDesignationInfo.controls.designation"
+    [interfaceOptions]="{ cssClass: 'custom-popover' }"
+    interface="interface"
```
MBuchalik: `interface="interface"` looks wrong to me :)

frontend/src/app/model/forms-data.model.ts
```diff
@@ -20,3 +21,10 @@ export interface ContactInfo {
   linkedIn: string;
   xing: string;
 }
+
+export interface FieldDesignationInfo {
+  // tslint:disable-next-line: no-any
+  field: string;
+  // tslint:disable-next-line: no-any
+  designation: string;
```
MBuchalik: Hm no need to disable tslint here, am I right?

frontend/src/theme/custom.scss
```diff
@@ -124,3 +124,25 @@ ion-item.item-has-focus {
...
+.custom-popover {
+  [aria-checked='true'].sc-ion-alert-ios .alert-checkbox-icon.sc-ion-alert-ios,
+  [aria-checked='true'].sc-ion-alert-md .alert-checkbox-icon.sc-ion-alert-md {
+    border-color: var(--ion-color-tertiary);
+    background-color: var(--ion-color-contrast);
+  }
+
+  .alert-checkbox-label.sc-ion-alert-ios,
+  .alert-checkbox-label.sc-ion-alert-md {
+    //font-size: 18px;
```
MBuchalik: Maybe, remove that line. (I mean the commented out one)

frontend/src/app/forms/steps/field-designation-page/field-designation-page.component.html
```diff
@@ -0,0 +1,41 @@
+<div class="main-description">
+  <h1>Industries and Positions</h1>
```
MBuchalik, line 2: What do you think - should we maybe use the same title as in the right menu? Otherwise, it might be confusing to the user.

MBuchalik (commented):
All in all, really good stuff! 👍 Clean code, working perfectly.

I left a few comments, but they are mostly related to some really minor things.

I would recommend not to merge this now, since the typization is currently wrong. (When looking at the interface: You have specified this as a `string` and not as `string[]`). It is currently simply not possible to use `string[]` due to the issue we talked about in Slack. So, I'd recommend to wait until we have fixed the issue and then to incorporate the necessary changes.

In case you are wondering why it still compiles: In the template, you can assign pretty much "anything" to the ion-select element. It doesn't check whether we expect a `string` or `string[]` as an output. (Since the typesafe implementation is a thing we built on our own...). So, it compiles, and under the hood still pushes a `string[]` into the field that was actually marked as `string`.

frontend/src/app/common/forms/forms.ts
Bharathravi1995: Hey Martin this was the work around code I did for to proceed with. Will be removed once your changes are merged.

frontend/src/app/forms/steps/field-designation-page/field-designation-page.component.html
Bharathravi1995: Just leave without interface which will be alert by default.

frontend/src/app/model/forms-data.model.ts
Bharathravi1995: This was added before I discussed with you. Once your changes are merged this will become Array<string>

frontend/src/app/forms/steps/field-designation-page/field-designation-page.component.html
denoribi: Okay i will leave this and change the name of the class.

frontend/src/app/forms/steps/field-designation-page/field-designation-page.component.html
denoribi: I can do this.

frontend/src/app/forms/steps/field-designation-page/field-designation-page.component.html
denoribi: Yep this is an error. But on this error i could now see, that the interface="popover" was unnecessary here too, because both popover dropmenus looked the same. :) I delete this line too.

frontend/src/theme/custom.scss
denoribi: Okay, that was rest of some testing the designs, but i didn't liked it.

frontend/src/app/forms/steps/field-designation-page/field-designation-page.component.html
denoribi: This is a good idea, but i was not sure if "field and designation" or "Industries and Position" is the better name and when we translate this later in german, because i thought this app should be in german, how will we name it then?

MBuchalik: Hm I like "Industries and Position" or "Industries and Positions" (you can select more than one position, hence the plural). 

Regadring the German version: How about directly translating it to "Industrien und Positionen" or sth like that?

denoribi:
@Bharathravi1995 sorry i had some merge conflicts, so i needed to override your last commit.

MBuchalik:
LGTM. From my point if view, you can merge it :)

Bharathravi1995:
> @Bharathravi1995 sorry i had some merge conflicts, so i needed to override your last commit.

Oh thats bad. We should avoid this kind of issues in future as commits are considered for grading. Also try to avoid git push --force .

MBuchalik:
Yeah, that was kinda my bad. I helped him get rid of the merge conflict using a rebase... That just killed your last commit. Sorry :/ 

(I am not entirely sure why your commit vanished entirely. That usually shouldn't happen even with a rebase if I am not mistaken...)

MBuchalik:
Anyways - feel free to merge it once you are done :)

PR #51, opened 2020-06-05
6 pictures

denoribi:
I have first Implemented basic skeleton for picture Entry and the functionality to take pictures. 

Then @dalex135 and me added the sliding and deletion function of the pictures and changed some designs, so it looks more like in the mockups.

I did change the picture so store in a string in a FormControl. With this i could fix th problem with switching the pages and then losing the picture.


frontend/package.json
```diff
@@ -30,6 +30,7 @@
     "@ionic-native/splash-screen": "^5.24.0",
     "@ionic-native/status-bar": "^5.24.0",
     "@ionic/angular": "^5.1.0",
+    "@ionic/pwa-elements": "^1.5.2",
```
MBuchalik, line 33: Please don't forget to add it to the Bill of Materials once this PR has been accepted. :)

frontend/src/app/forms/forms.module.ts
```diff
@@ -34,6 +35,7 @@ import { SubmitPageComponent } from './steps/submit-page/submit-page.component';
     EducationInfoComponent,
     EducationInfoEntryComponent,
     FieldDesignationPageComponent,
+    PicturesEntryComponent,
```
MBuchalik: Wouldn't it be better to use the singular form ("Picture") instead of the plural ("Pictures")? We only allow the user to take exactly one picture.

Also: How about renaming "PictureEntry" to "Profile Picture" or "Picture" or similar?

MBuchalik: If you decide to change this name, please don't forget to change it in all other places as well :)

frontend/src/app/forms/forms.page.ts
```diff
@@ -72,6 +73,9 @@ export class FormsPage implements OnInit, OnDestroy {
       field: new FormControl<string[]>([], Validators.required),
       designation: new FormControl<string[]>([], Validators.required),
     }),
+    pictureEntry: new FormGroup<PictureEntry>({
+      picString: new FormControl('', Validators.required),
```
MBuchalik: Should we maybe rename this to "pictureBase64" or something like this?

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.html
```diff
@@ -0,0 +1,45 @@
...
+  <div
+    *ngIf="
+      formsData.controls.pictureEntry.controls.picString.invalid;
+      else elseBlock
+    "
+  >
+    <span>
+      Add a picture of yourself by tapping on the image icon.
+    </span>
+    <ion-button
```
Bharathravi1995: On clicking on x mark on the top left corner uncaught error is thrown. Can you check that.

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.ts
```diff
@@ -0,0 +1,47 @@
...
+      allowEditing: false,
+      resultType: CameraResultType.Base64,
+      source: CameraSource.Camera,
+    });
+
+    /*this.photo = this.sanitizer.bypassSecurityTrustResourceUrl(
+      image && image.dataUrl
+    );
+    */
+    this.photo = 'data:image/jpeg;base64,' + image.base64String;
```
Bharathravi1995: Is this variable used any where? I couldn't able to spot it.

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.html
```diff
@@ -0,0 +1,45 @@
+<div class="main-description">
+  <h1>Picture</h1>
+
+  <!--<ion-img [src]="photo"></ion-img>-->
```
MBuchalik: Please remove the commented-out line :)

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.html
```diff
@@ -0,0 +1,45 @@
...
+        </ion-item-option>
+      </ion-item-options>
+      <ion-item class="ion-no-padding">
+        <ion-img
+          [src]="formsData.controls.pictureEntry.controls.picString.value"
+        ></ion-img>
+      </ion-item>
+    </ion-item-sliding>
+  </ng-template>
+</div>
```
MBuchalik: You should really only put a description into "main description", and not the whole content. 

Also: I believe you could improve the general style by using something like the following

``` html
  <button (click)="takePicture()" class="custom-button">
    <ion-icon name="person"></ion-icon>
  </button>
```
and in the css, sth like

``` css
.custom-button {
  border-radius: 100%;
  background: var(--ion-color-primary);
  height: 150px;
  width: 150px;
  margin-top: 3rem;
  margin-right: auto;
  margin-left: auto;
  display: flex;
  align-items: center;
  justify-content: center;
  color: var(--ion-color-primary-contrast);
  font-size: 60px;
  outline: none;
}
```

(The css is pretty unstructured - this is really just an example tho :) )

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.html
```diff
@@ -0,0 +1,45 @@
...
+    >
+      <ion-icon name="person-circle"></ion-icon>
+    </ion-button>
+  </div>
+  <ng-template #elseBlock>
+    <span>
+      Good job! If you want to take another picture, just delete the actual
+      picture by sliding on it and take another one.
+    </span>
+    <ion-item-sliding class="ion-no-padding custom-sliding">
```
MBuchalik: Instead of sliding the item to delete it, maybe just show a button? Sliding such a large item to delete is pretty uncommon I guess... I would just show a "delete" button somewhere. Maybe just a trash icon.

What do you think?

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.scss
```diff
@@ -0,0 +1,15 @@
+.custom-button {
+  flex-grow: 1;
+  flex-direction: column;
+  align-items: center;
+  display: flex;
+  justify-content: flex-end;
```
Bharathravi1995: Why do we need this as flex-end aligns the content at end? Also flex-grow: 1;
  flex-direction: column;
  align-items: center doesn't really needed now as there is only one element inside it.

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.html
```diff
@@ -0,0 +1,45 @@
...
+          expandable
+          (click)="deletePictureEntry()"
+        >
+          <ion-icon name="trash" size="large"></ion-icon>
+        </ion-item-option>
+      </ion-item-options>
+      <ion-item class="ion-no-padding">
+        <ion-img
+          [src]="formsData.controls.pictureEntry.controls.picString.value"
+        ></ion-img>
```
MBuchalik: Is there any advantage when using ion-img instead of a simple img?

frontend/package.json
Bharathravi1995, line 33: Add it in BOM

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.ts
```diff
@@ -0,0 +1,47 @@
...
+
+@Component({
+  selector: 'form-pictures-entry',
+  templateUrl: './pictures-entry.component.html',
+  styleUrls: ['./pictures-entry.component.scss'],
+})
+
+/*
+  To save Files as Data.url just change restultType and uncomment everything in this file
+*/
```
MBuchalik: Please just remove all commented-out things :)

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.ts
```diff
@@ -0,0 +1,47 @@
...
+export class PicturesEntryComponent {
+  constructor(/*private sanitizer: DomSanitizer*/) {}
+  @Input()
+  formsData: FormGroup<FormsData>;
+
+  photo: SafeResourceUrl;
```
MBuchalik: This property is currently not needed, am I right? (I mean the `photo` property)

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.ts
```diff
@@ -0,0 +1,47 @@
...
+
+  async takePicture(): Promise<void> {
+    const image = await Plugins.Camera.getPhoto({
+      quality: 100,
+      allowEditing: false,
```
MBuchalik: I am curious - what does "allowEditing" do if we enable it? It probably doesn't do anything when previewing in the browser. But in our real app, would this display some sort of cropping controls?

MBuchalik (commented):
All in all, a pretty solid enhancement! Nice progress 👍  I added a few comments.

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.html
MBuchalik: I know that this is not part of my mockups. Would it help if I added something like this?

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.ts
```diff
@@ -0,0 +1,47 @@
...
+      resultType: CameraResultType.Base64,
```
Bharathravi1995: I have a doubt. In networks, I could able to see blob object being created even though resultType is set as Base 64. I think it is generated by the capacitor for image preview. Is it the general behaviour? Irrespective of the type blob will be created? If  so will it also be revoked automatically? I hope it should be just because it was created by capacitor. 

Also will there be any configuration we can have for image size? Will always same size  image be returned? Or do we need to have any max width in CSS to restrict the size.

Bharathravi1995 (commented):
@denoribi @dalex135  Good PR in general. I have commented some of my doubts I have with respect to blob.

MBuchalik:
I had a look at the acceptance criteria. Here, among others, it says "it should directly open the front camera". I believe this is currently not fulfilled. First of all, on a native device, it might show a prompt first. So you should set the `source` attribute to Camera. Also, I believe it will open the rear camera by default. So you should maybe set this to Front.

frontend/src/app/forms/forms.module.ts
denoribi: yep changed it in "profile picture"

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.html
denoribi: yeah a little mockup where trash button should then be can help.
And yeah maybe sliding is for the picture not rly goo here.

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.html
denoribi: I just used it because its from ionic. So should we better use just img? it works the same in this case @MBuchalik

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.ts
denoribi: This is just if we want to use Data.url but, as we decided to go for base64string i will delete this.

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.ts
denoribi: I didn't test it, but i did read that in the mobile phone, you should then be able to edit the photo after you have taken it. Which edit functions we then have, needs to be tested. Also there is a correlation, if you want to save the picture in the photo gallery of the mobile phone. For this i think you need to allow editing then.

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.html
MBuchalik: I have now created the mockups for it. Just click on the round button in the middle in order to see the changes.

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.html
denoribi: tried to solve it with a catch block, but this actually didn't work. Didn't find a fix to this, but i'm not sure if this is a problem. I will need to do some more researches on this.

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.ts
denoribi: no, it was there just for using dataUrl, but we decide not to use it. I just removed it.

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.html
MBuchalik: I think we can just use a normal img here. If I understood it correctly in their docs, ion-img is rather a thing that can be used for large lists of images.

(But honestly, there shouldn't be a huge difference in our case. So if you want to keep ion-img, then that's fine as well.)

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.ts
MBuchalik: >Also there is a correlation, if you want to save the picture in the photo gallery of the mobile phone. For this i think you need to allow editing then.

In the Capacitor docs, it says 

> 
> // Whether to save the photo to the gallery. If the photo was picked from the gallery, it will only be saved if edited. Default: false
> saveToGallery ?: boolean;

I think that rather means: If you allow users to pick a photo **from** the gallery. Then the edited photo will only be saved again to the gallery if some changes were made to it. So this doesn't apply to our use case imho.

In general: Saving the photo to the gallery isn't something we should do (data protection concerns, etc.). But it would be cool if editing would be possible. So, maybe, we could check that out once we have actually built the app for iOS. :)

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.ts
denoribi: I will do some research about this.

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.html
denoribi: Yes this worked better for me, just did some smaller changes on this.

denoribi:
Changed it so the front camera should be used, but i would need to test this on the phone then.

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.scss
denoribi: fixed this. (will upload it tomorrow morning)

frontend/src/app/forms/steps/pictures-entry/pictures-entry.component.ts
denoribi: Added the max-width, so picture gets always resized. ( this will be uploaded tomorrow)

denoribi:
Just changed the picture to be resized correct and change the trash button for better code. I will upload the commits tomorrow morning.

Bharathravi1995:
@denoribi Once You are done with the changes. Either you or @MBuchalik  can merge this.

MBuchalik:
LGTM 👍

PR #52, opened 2020-06-06
Implement Key Competencies step

MBuchalik:
I have implemented the frontend for the Key Competencies step. (Originally, this was called "Technical Knowledge".)

This PR is pretty huge. I'd suggest to simply try the functionality out locally.


frontend/src/app/forms/forms.page.ts
```diff
@@ -72,6 +73,12 @@ export class FormsPage implements OnInit, OnDestroy {
       field: new FormControl<string[]>([], Validators.required),
       designation: new FormControl<string[]>([], Validators.required),
     }),
+    keyCompetencies: new FormGroup<KeyCompetencies>({
+      languages: new FormControl([], Validators.required),
+      professionalSoftware: new FormControl([]),
+      databases: new FormControl([]),
+      programmingLanguagesAndFrameworks: new FormControl([]),
+    }),
```
denoribi, line 81: For now its okay to just use only languages as validator, but maybe we can make this dependent, on which jobs the student is applying to. So as an dev, he should need to fill out the programming languages too.

frontend/src/app/forms/steps/key-competencies/key-competencies.component.scss
```diff
@@ -0,0 +1,13 @@
+.list-item-label {
+  font-weight: 500;
+  letter-spacing: 1px;
+  font-size: 1.2rem;
+  white-space: normal;
+}
```
denoribi, line 6: I think the font-weight is here a bit too strong, 400 would be enough for me.

frontend/src/app/forms/steps/key-competencies/rating-modal/item-popover/item-popover.component.html
```diff
@@ -0,0 +1,37 @@
...
+      </ion-button>
+    </ion-buttons>
+  </ion-toolbar>
+</ion-header>
+
+<ion-content class="ion-padding">
+  <div class="description">
+    Please rate your skills from 1 (not good) to 5 (excellent).
+  </div>
+
```
denoribi, line 25: I don't think they should vote from "not good" to "excellent". I mean as an applicant, I never want to say, I'm not good in something. I have experience or i don't have one and then i will just not click on it and say, i have skills in it. So here would be instead of "not good" something like "basic" better, or maybe for the different key competencies different rating scales.

frontend/src/app/forms/steps/key-competencies/rating-modal/rating-modal.component.html
```diff
@@ -0,0 +1,79 @@
+<ion-header>
+  <ion-toolbar>
+    <ion-buttons slot="end">
+      <ion-button (click)="closeModal()">Close</ion-button>
+    </ion-buttons>
+  </ion-toolbar>
```
denoribi, line 6: Would it here be better to say "Save" or "rate" instead of "Close" ?

denoribi (commented):
Really good job Martin! 
I mainly focused on the design in my review, so the most comments are just some proposals for design changes. 

But overall for me it looks really good and i think this could be merged. Maybe @Bharathravi1995 wants to have a look on it too, before, because its really a big US.

frontend/src/app/forms/steps/key-competencies/key-competencies.component.scss
MBuchalik: I wanted to make it like in the mockups. So if you are ok with it, I would leave it as it is for now :) The items are pretty important, so I guess it is good to have them stick out a bit

frontend/src/app/forms/steps/key-competencies/rating-modal/item-popover/item-popover.component.html
MBuchalik: Hm what about the following: We leave it as it is for now and then discuss in the team on how we should call it. Or even ask the Industry Partners. Would that be ok?

frontend/src/app/forms/steps/key-competencies/rating-modal/rating-modal.component.html
MBuchalik: Currently, when using the slider to update the rating, then the rating is automatically saved. If there is a "Save" button, then as a user, I'd expect that nothing is saved until I press this button. Because of this, I decided to call it "Close". What do you think - is this ok?

In the original mockups, I had such a feature: There was one "Save" and one "Cancel" button. But while implementing this, I figured that having just one "Close" button makes things easier to implement.

denoribi: But when I think of "Close", then I'm not sure if i the data is now saved and transposed. So i thought about something like "Submit" then maybe. But we can ask the team about this, too. :)

PR #53, opened 2020-06-07
Key Competencies: Backend

MBuchalik:
I have integrated fields into the backend to store the Key Competencies. The API itself is probably not ideal, but I guess it is ok for now (for testing). We can/should maybe revamp this while actually "connecting" the frontend with the backend.


Backend/src/main/java/Digitact/Backend/Model/User/Applicant.java
```diff
@@ -74,4 +80,10 @@ public Positions getPositions() {
     public void setPositions(Positions positions) {
         this.positions = positions;
     }
+
+    public void addKeyCompetence(KeyCompetence keyCompetence) { keyCompetencies.add(keyCompetence); }
+
+    public List<KeyCompetence> getKeyCompetencies() { return new ArrayList<KeyCompetence>(keyCompetencies); }
+
+    public void addKeyCompetencies(List<KeyCompetence> keyCompetencies) { keyCompetencies.forEach(comp -> this.keyCompetencies.add(comp)); }
```
vishwasanavatti: Its always good to maintain consistency in method naming. So, it would be better to change "addkey" to "setKey".

vishwasanavatti (commented):
Everything looks fine except for that naming convention. Haven't tested locally but looks like it works.

Backend/src/main/java/Digitact/Backend/Model/User/Applicant.java
MBuchalik: To me, a method named "set" always sounds like it **overrides** the current content with what is being passed as parameter. This is not the case here: It appends stuff, so the old content is not being removed. Because of this, I wasn't really comfortable with using "set". What do you think about this?

I guess we could do one of the following two things:
1. I just rename it to "set". And we keep in mind that we should refactor this later.
2. I rename all "set..." methods to "add..." or "append..." or whatever.

What would you prefer?

Nasser-Eddin-Nasser: The word set in OOP means to declare&change a value to a variable, while add means to insert an element to a list for example. 
Usually, "add" used to add an object to another object. 
For example: 
"SetList" means, myList= newList; // old list is removed
"AddList" means myList += newList; // extend my old list  .

Nasser-Eddin-Nasser: FYI: 
usually, the "key" should be final, which means its only set by initialization through the constructor. so we don't need a set method. :)

Nasser-Eddin-Nasser: I'll review the code and refactor it today!

MBuchalik: > The word set in OOP means to declare&change a value to a variable, while add means to insert an element to a list for example.

OK, so my implementation should be fine, right? If not, then please elaborate a bit more on what I should change and in which way. :) 

> usually, the "key" should be final

Which key are you referring to?

> I'll review the code and refactor it today!

Could we merge my PR before you refactor anything? Otherwise, I will get merge conflicts on that one.

Backend/README.md
```diff
@@ -118,6 +118,13 @@ If you want to access the database, you are most likely prompted for a password.
       "grade": double,
       "graduation_date": String
     }
+  ],
+  "keyCompetencies": [
+    {
+      "category: KeyCompetenciesCategory,
```
Nasser-Eddin-Nasser: fine **Syntax** error by "category: KeyCompetenciesCategory,

Backend/src/main/java/Digitact/Backend/Model/User/Applicant.java
Nasser-Eddin-Nasser: I'll check the code and fix the "bugs/errors" if there is and will let u know what I changed and why @MBuchalik  :)

Backend/README.md
MBuchalik: You mean the missing quote after `"category`?

Backend/src/main/java/Digitact/Backend/Model/User/Applicant.java
MBuchalik: I didn't even know you have write access on my repo :D

Backend/README.md
Nasser-Eddin-Nasser: yeah, and KeyCompetenciesCategory should be declared

Nasser-Eddin-Nasser:
I pulled and ran the code locally, it works perfectly, It can be merged with master  :)

PR #55, opened 2020-06-08
Finalization of Applications

vishwasanavatti:
This PR has 3 new pages and database changes.
1) Impression page - impression to be entered by HR staff.
2) Finalize page - to finalize the application after all the details have been filled and mark it ready for sending it to HR monitor.
3) Applications page - Here the not finalized list of applications is listed and also shows the count of applications that are finalized and has send button.


frontend/src/app/applications/applications.page.html
```diff
@@ -0,0 +1,72 @@
...
+      <div class="content">
+        <div class="main-description info-content">
+          <h1>Not Finished</h1>
+          This list includes all the applications that have been submitted by
+          the applicant but not been rated by HR staff.
+        </div>
+      </div>
+    </div>
+
+    <div style="padding: 1.5rem; padding-top: 0.5rem !important;">
```
Bharathravi1995: Can you change it to class

frontend/package.json
```diff
@@ -37,7 +37,7 @@
     "zone.js": "~0.10.2"
   },
   "devDependencies": {
-    "@angular-devkit/build-angular": "~0.901.4",
+    "@angular-devkit/build-angular": "^0.901.7",
```
MBuchalik: Why was this change needed? (Especially: Why the change from `~` to `^`?)

frontend/src/app/forms/model/steps.model.ts
```diff
@@ -12,7 +12,15 @@ export enum ApplicationStep {
...
+  ImpressionInformation = 'impression-information',
+  Finalize = 'finalize',
+}
+
 /**
  * All of the possible steps as an array.
  */
 export const ApplicationStepsArr = Object.values(ApplicationStep);
+
+export const hrRatingStepArr = Object.values(hrRatingStep);
```
MBuchalik: I think it would be better if we put the HR-part (hrRatingStep and hrRatingStepArr) into app/rating/model/. Maybe, we could create a file "app/rating/model/steps.model.ts".

frontend/src/app/forms/side-menu/side-menu.component.html
```diff
@@ -32,21 +32,20 @@ <h1>Progress</h1>
...
           [ngTemplateOutletContext]="{
-            step: APPLICATION_STEP.EducationInformation,
-            displayName: 'Education Information',
-            isValid: formsData.controls.educationInfo.valid,
-            isActive: currentStep === APPLICATION_STEP.EducationInformation
+            step: APPLICATION_STEP.FieldDesignationPreference,
+            displayName: 'Industries and Positions',
+            isValid: formsData.controls.fieldDesignationInfo.valid,
+            isActive:
+              currentStep === APPLICATION_STEP.FieldDesignationPreference
```
MBuchalik: Oh yeah, I saw this as well :D (Fixed with my PR)

frontend/package.json
Bharathravi1995: I think ~0.901.4 itself can fetch minor patch version .7

frontend/src/app/home/home.page.html
```diff
@@ -1,3 +1,26 @@
+<ion-header class="ion-no-border ion-padding">
+  <ion-toolbar>
+    <ion-buttons slot="end">
+      <ion-button
+        id="notification-button"
```
MBuchalik: In Angular, you almost never use an ID in the HTML. Please just use a class for this.

MBuchalik: The same applies for the "notification badge".

frontend/src/app/home/home.page.html
```diff
@@ -1,3 +1,26 @@
...
+        style="min-height: 100px;"
```
Bharathravi1995: Can u remove inline style

frontend/src/app/home/home.page.scss
```diff
@@ -42,3 +42,22 @@ h3 {
 .logo-square {
   fill: var(--ion-color-secondary);
 }
+
+.custom-icon {
+  font-size: 64px;
+  color: lightgray;
```
MBuchalik: Please use Ionics color variables for this. I think "ion-color-light-shade" could be the one for this.

Also, this button is waaay to large :D  I'd suggest to set the font-size to 40px. On the button itself, you can also set the height to 40px (or maybe a bit more, I haven't tested that)

frontend/src/app/home/home.page.scss
```diff
@@ -42,3 +42,22 @@ h3 {
...
+}
+
+#notification-button {
+  position: relative;
+  overflow: visible !important;
+}
+
+#notifications-badge {
+  background-color: red;
```
MBuchalik: Please use the Ionic color variables here (`var(--ion-color-secondary)`).

frontend/src/app/rating/rating.page.ts
```diff
@@ -4,38 +4,280 @@
  *
  */
 
-import { Component } from '@angular/core';
+import { Component, OnDestroy, OnInit } from '@angular/core';
 import { Validators } from '@angular/forms';
+import { ActivatedRoute, Router } from '@angular/router';
 import { NavController } from '@ionic/angular';
+import { AlertController } from '@ionic/angular';
+import { ToastController } from '@ionic/angular';
```
MBuchalik: With the changes from my PR that have been merged a few minutes ago, the iimports for AlertController and ToastController are blacklisted. Please use the ones I added into common/ion-wrappers/ instead. :)

(At the time you wrote this, my changes weren't merged into master. So you couldn't have known that.)

frontend/src/app/rating/rating.page.ts
```diff
@@ -4,38 +4,280 @@
...
     this.navController.navigateBack('/home');
   }
+  /**
+   * In this method popup is displayed to confirm finalize the applicant information
+   */
+  async finalize(): Promise<void> {
+    const alert = await this.alertController.create({
+      header: 'Finalize',
+      message:
+        'Do you really want to finalize this form? it wont be possible to edit later.',
```
MBuchalik: There is a typo: "wont" instead of "won't" :) And the "it" should be uppercase.

frontend/src/app/services/storage-handler.service.ts
```diff
@@ -6,100 +6,132 @@ import { Injectable } from '@angular/core';
...
 import { FormsData } from '../model/forms-data.model';
+import { RatingForm } from '../rating/model/rating-form.model';
 
 @Injectable({
   providedIn: 'root',
 })
 export class StorageHandlerService {
-  constructor(private storage: Storage) {}
+  applicantDetailsDb: Storage;
+  apllicantRatingsDb: Storage;
```
MBuchalik: There is a typo: "apllicant" instead of "applicant"

frontend/src/app/services/storage-handler.service.ts
```diff
@@ -6,100 +6,132 @@ import { Injectable } from '@angular/core';
...
+    this.apllicantRatingsDb = new Storage({
+      name: 'digitact',
+      storeName: 'applicants-rating',
+      driverOrder: ['indexeddb', 'localstorage'],
+    });
+    this.commonPropertiesDb = new Storage({
+      name: 'digitact',
+      storeName: 'common-properties',
+      driverOrder: ['indexeddb', 'localstorage'],
+    });
```
MBuchalik: I am happy that it worked like that. :) We definitely shouldn't forget to add the sqlite storage as well

frontend/src/app/services/storage-handler.service.ts
```diff
@@ -6,100 +6,132 @@ import { Injectable } from '@angular/core';
...
+  }
+
+  /**
+   * Used to generate next Id
+   * @returns Promise of string containing next id
+   */
+
+  async getNextId(): Promise<string> {
+    const val = await this.commonPropertiesDb.get('recentId');
```
MBuchalik: Should we maybe rename "recentId" to "recentApplicationId" or "recentApplicantId" or "lastApplicationId"?

frontend/src/theme/custom.scss
```diff
@@ -145,3 +145,20 @@ ion-item.item-has-focus {
...
+ * Customization of the alert in the dropdownmenu of the finalize-page
+*/
+.custom-class {
+  .cancel-button {
+    color: blue;
+  }
+  .finalize-button {
+    color: red;
+  }
+}
```
MBuchalik: Please don't hard-code these colors here. If you want to use a red color, then use `var(--ion-color-warning)` (or the secondary color, which is at least in our case red as well).

Also, maybe make everything a bit more generic: Just create the css classes like so:

``` css
.custom-alert-button-colors {
  .color-secondary {
    color: var(--ion-color-secondary);
  }
}
```

And finally: I know that in the mockups, one button is blue. But I think it would be ok to keep the default color.

frontend/src/theme/custom.scss
```diff
@@ -145,3 +145,20 @@ ion-item.item-has-focus {
...
+
+.ok-button {
+  button {
+    color: blue !important;
+  }
+}
```
MBuchalik: I think, if you apply the changes of my previous comment, then you can just delete this.

frontend/src/app/applications/applications.page.ts
```diff
@@ -0,0 +1,67 @@
...
+
+    this.notFinishedEntry = inp.filter((x) => x.isRated === 0);
+
+    this.countNotFinalized = this.notFinishedEntry.length;
+  }
+  /**
+   * In this method navigation to HR rating is handled.
+   */
+  goToRating(id: string): void {
+    this.navController.navigateForward(['/rating'], { state: { id } });
```
MBuchalik: Just out of curiosity: Is there any reason why you put the ID into the state, instead of just using it directly in the URL? The id is, after all, just a string.

MBuchalik (commented):
Wow, a pretty large PR. I tested it locally and it seems to work fine! 🎉 I added a few comments.

frontend/src/app/forms/forms.page.ts
```diff
@@ -205,15 +212,23 @@ export class FormsPage implements OnInit, OnDestroy {
...
+    });
+    this.formsData.controls.submittedTime.setValue(time);
+
+    this.storage.getNextId().then((key) => {
+      this.formsData.controls.id.setValue(key);
+
+      this.storage.addItem<FormsData>(
+        this.storage.applicantDetailsDb,
+        key,
+        this.formsData.getRawValue()
```
MBuchalik, line 239: In 99.9% of the cases, you don't actually want to use "getRawValue". Why is this necessary here? Please keep in mind that getRawValue will include all fields, including disabled ones. This quickly leads to bugs. Or it "leaks" data the user didn't even want to submit. So, one should be really really careful when using this.

frontend/src/app/services/storage-handler.service.ts
Bharathravi1995: SQLite is next priority.

frontend/src/app/forms/forms.page.ts
Bharathravi1995: Ya I know what it is being used for and aware of it. The main reason is I have added id and isRated flag in the interface , I don't want these to validated as it is system generated. SO I disabled these fileds. But I want these fields in storage. Hence used getRawValue.

frontend/src/app/forms/model/steps.model.ts
vishwasanavatti: ya but why create another file. I felt its unnecessary to create another one but will create it if its really required to be separated.

frontend/src/app/home/home.page.html
vishwasanavatti: ok will change that.

frontend/src/app/home/home.page.scss
vishwasanavatti: ya ok. I saw somewhere that icon size should be multiples of 2^n so gave it 64 as 32 was too small. Will check with your suggestion.

frontend/src/app/forms/forms.page.ts
MBuchalik: > I don't want these to validated as it is system generated

If you don't add a Validator, then they won't be validated. ;) So I see no problem here...

The only potential issue I see at this point is that the progress bar would also "react" to these values. But we should rather fix the progress bar instead of doing sth like this.

Is there any other reason why this is necessary? If not, then I strongly recommend against using `getRawValue()` (as described in my first comment).

frontend/src/app/forms/model/steps.model.ts
MBuchalik: The HR steps don't have anything to do with the "main form". I think you should rather think like "what belongs together" instead of "how can I use less files" ;) 

The HR steps are a completely separate thing. They even live in a separate module. And you should usually not include files over the boundaries of modules. (Of course, there are exceptions to this rule...)

frontend/src/app/applications/applications.page.ts
Bharathravi1995: Ya it can be implemented as dynamic route as well. But I didn't feel the necessity to implement it, as just passing id in state solves the needs. Also In future if we want some other data, we can send it. Otherwise an extra step of querying record is needed.

frontend/src/app/home/home.page.scss
MBuchalik: Hm I have never heard of this rule

frontend/src/app/forms/forms.page.ts
Bharathravi1995: Currently the interface is `export interface FormsData {
  id: string;
  isRated: number;
  submittedTime: string;
  basicInfo: BasicInfo;
  contactInfo: ContactInfo;
  educationInfo: EducationInfo;
  fieldDesignationInfo: FieldDesignationInfo;
}`

Id and isRated are not under FormGroup as like other fields. And I don't think it is correct to move as a Group. If you leave without Validators.required valid property becomes true. We are using value property of the first level Control object to calculate progress. If we did not disable it I have to make poor hack in the code to work. So I used disable() to remove it from control as suggested in https://angular.io/api/forms/AbstractControl#disable . But I need the disabed values in storage. hence I used getRawValues(). Also that is one of the usecase why  such methods exists.

MBuchalik: > We are using value property of the first level Control object to calculate progress.

Then we should definitely refactor how the progress is calculated. 

>  If we did not disable it I have to make poor hack in the code to work.

Not necessarily. We can simply refactor the method that updates the progress value. One idea: Have an object that, for each top-level element, tells us if it should be part of the progress counter or not. (This would also be super handy when we add pages that only include optional items in the future.)  
Is there anything speaking against this?

>  So I used disable() to remove it from control as suggested in https://angular.io/api/forms/AbstractControl#disable 

I think you have mis-understood the wordings used there. It says:
> Disables the control. This means the control is exempt from validation checks and excluded from the aggregate value of any parent. Its status is DISABLED.

This is really much more than just saying "I don't want this control not to be validated". In general, you really only use this if you want to "grey out" a form element if the user is not supposed to enter data into this for now. (Of course, this is just one example of a pretty common use case.)

> Also that is one of the usecase why such methods exists.

We should avoid using such methods at all costs. I believe such methods exist to handle edge cases that you really cannot handle in another way. Or for debugging purposes. But in our case, we can really just update our progress counter algorithm and that's it.

Bharathravi1995: Ya I agree disable() is more than avoiding validation. It will remove from the control itself which is the one I want to achieve. Also I feel the right way to do. Why id and isRated which are not filled by user to exist under control. For now it works without any hack. Adding flag at top level should work. But I don't think of a necessity now. We can refactor later on when we implement such pages.

frontend/src/app/forms/model/steps.model.ts
vishwasanavatti: ok will create a new file.

frontend/src/app/forms/forms.page.ts
MBuchalik: >For now it works without any hack. 

Using `getRawValue()` for this is a hack. It is just not the way you deal with form values. It is error-prone and may leak data: Other developers might think that their disabled entries will not be part of the final result. And this will then accidentally still be submitted to the server. By using getRawValue, you also simply disallow any developer from using the disable function in any part of the form in the way they are used to: It is simply not possible anymore to use "disable" on a form element in order to actually hide data from the final result.

But I got the feeling it makes no sense to argue here anymore. Just merge it like it is, but please at least add a prominent comment/info about the fact that disabled form controls are still submitted. This is not how forms are usually dealt with and any future developer on this code base may fall for this.

Btw: Another simple solution could have been to use a structure like this:

``` ts
interface MyData {
  id: number;
  isRated: boolean;
  submittedTime: Date;

  formValue: FormValue<FormsData>;
}
```

Then, when submitting/storing the data, you could simply have done the following:

``` ts
const data: MyData = {
  id: theId,
  isRated: false,
  submittedTime: new Date(),

  formValue: this.formsData.value;
}

// The following is simplified...
this.storage.addItem<MyData>(data);
```

Bharathravi1995: A developer should be aware of what he is disabling. A need of disabling arrives when he doesn't want a field to be controlled. When he doesn't want to control? When it is not user entered fields. Also developer should refrain from adding unwanted fields in the interface other than some thing used for state maintainence in client  and system generated. These data structure are meant only for  local storage and should not be sent to the server directly. I don't think there will be a data leak if interface is properly handled. Let's say some fields has to enabled or disabled based on another formControl value. In this case if he does not want a field he can disable it. Here the default empty  value will be stored as done in the FormGroup instance creation for disabled fields. In this case if we use .value instead of getRawValues() the object will be no more of type in interface until we make it optional.

 Initially I thought of doing the same way as suggested in your interface, but it creates one more nested level of the object.

frontend/src/app/rating/rating.page.ts
vishwasanavatti: could not figure out how to make wont to won't. I tried with escape sequence but couldn't do it as lint fails.

MBuchalik: Can't you just use `const str = 'this won\'t be a problem'`? Or have we made the linter too strict? :D

MBuchalik:
If you are done with the changes suggested, feel free to merge this PR! 👍 
The only discussion that will remain open is https://github.com/Nasser-Eddin-Nasser/Digitact/pull/55#discussion_r436714326 but I guess we shouldn't get blocked by this. I still got the opinion that one should not use getRawValue() in the way it was done here.

Bharathravi1995:
> If you are done with the changes suggested, feel free to merge this PR! 👍
> The only discussion that will remain open is [#55 (comment)](https://github.com/Nasser-Eddin-Nasser/Digitact/pull/55#discussion_r436714326) but I guess we shouldn't get blocked by this. I still got the opinion that one should not use getRawValue() in the way it was done here.

Ya Martin, I have a different opinion about getRawValue() still. Anyways I will file an issue regarding it for tracking. I will think of it and will do the needful. Thanks for your suggestions and views on it. It has different perspective than mine.

frontend/src/app/forms/forms.page.ts
MBuchalik: > A need of disabling arrives when he doesn't want a field to be controlled. When he doesn't want to control? When it is not user entered fields.

If you don't want the user to enter anything, then just don't add it to the HTML.

There are use cases where you programatically fetch something and still want to have validators on that. Even if the actual input element is not directly seen on the UI.

> Also developer should refrain from adding unwanted fields in the interface other than some thing used for state maintainence in client and system generated.

What about the following use case? A checkbox asking "do you want to provide feedback to this (whatever)?". If it is checked, then some form items are displayed that the user now may fill out. But if the checkbox is not checked, then the user can just proceed. Do you think that, when the box is not checked, it really makes sense to include all of the disabled fields in the result?

From my point of view, this is a valid use case of using "disable()". 

Or, an example that is closer to our app: In our form we display a menu where the user may enter the schools and universities he graduated in. If an university is selected, then we ask for the name and, let's say, the grade of the final thesis. But if a school is selected, we only ask for the name. This is again one thing you would implement using the "disable()" function.

> Let's say some fields has to enabled or disabled based on another formControl value. In this case if he does not want a field he can disable it. Here the default empty value will be stored as done in the FormGroup instance creation for disabled fields.

The more elegant way is to simply not store the fields at all.


Finally, let's think about how this is handled in traditional HTML forms. If you wanted to store some data hidden from the user, then you would just create an input element with `type="hidden"`. In Angular, you simply don't create the HTML part of this form.

In traditional HTML forms, if you diabled an input field, it would not be sent to the server. In Angular, if you use "value", then the disabled inputs are not included. Coincidence?

The only reason why you are using these disabled fields in this PR is because of the progress indicator. If it didn't exist, then you wouldn't have had the need to disable the fields. So, from my point of view, one should simply fix the progress indicator instead.


But, I guess we should just keep going instead of wasting too much time on this discussion. You heard my points, I heard yours. I still would never ever use getRawValue(), but I cannot stop anybody from using it. :D

PR #56, opened 2020-06-08
Auto increment Id added

Bharathravi1995:


frontend/src/app/rating/rating.page.ts
```diff
@@ -247,7 +247,7 @@ export class RatingPage implements OnDestroy, OnInit {
         this.storage.addItem<RatingForm>(
           this.storage.apllicantRatingsDb,
           this.ratingForm.controls.id.value,
-          this.ratingForm.value
+          this.ratingForm.getRawValue()
```
MBuchalik, line 250: In 99.9% of the cases, you don't actually want to use "getRawValue". Why is this necessary here? Please keep in mind that getRawValue will include all fields, including disabled ones. This quickly leads to bugs. Or it "leaks" data the user didn't even want to submit. So, one should be really really careful when using this.

PR #58, opened 2020-06-08
npm update

denoribi:
I just did a "npm update" on my new laptop. I know its not necessary needed, but i think its nicer like this.


frontend/package.json
```diff
@@ -37,25 +37,25 @@
...
-    "@angular-devkit/build-angular": "~0.901.4",
-    "@angular/cli": "~9.1.4",
-    "@angular/compiler": "~9.1.4",
-    "@angular/compiler-cli": "~9.1.4",
-    "@angular/language-service": "~9.1.4",
+    "@angular-devkit/build-angular": "^0.901.7",
+    "@angular/cli": "^9.1.7",
+    "@angular/compiler": "^9.1.9",
+    "@angular/compiler-cli": "^9.1.9",
+    "@angular/language-service": "^9.1.9",
```
MBuchalik: I am wondering why npm update changed this from `~` to `^`. I think it would be better, for all entries, to have just the patch versions applied... Could you maybe manually change the entries back to be prefixed with `~`?

MBuchalik:
Yeah, I also like keeping my package.json up-to-date, even though from a technical point of view, this is not really necessary. 

Could you also update the Bill of Materials?

frontend/package.json
denoribi: yep i will change this tomorrow :)

denoribi:
yeah i will add them into the BOM after the changes are done and the pr is merged

MBuchalik:
LGTM 👍

PR #69, opened 2020-06-13
Document Selection

MBuchalik:
I implemented the Document Selection step.

I have not optimized it regarding the performance. Most importantly: Currently, all images are just full-size base64 strings. This could slow down the DOM. In my experiments, everything was ok, but you never know. So, in the future, we might think about optimizing this.

I also implemented a mechanism to make this step optional.

@denoribi Could you check this out? :)


MBuchalik:
Just did a rebase. For whatever reason, at least on my screen, the default merging option is "Squash and merge" now. Please make sure to use the normal "merge" instead when merging this PR.

frontend/src/app/forms/forms-routing.module.ts
```diff
@@ -19,6 +19,14 @@ const routes: Routes = [
         (m) => m.ConfirmationPageModule
       ),
   },
+  // Important: Do not route to the Image Viewer on your own. Instead, please use the provided Service for that.
+  {
+    path: 'image-viewer',
```
denoribi, line 24: So by manually routing to it, it doesn't work like you mentioned. Do we need an AuthGuard for this then too?

denoribi (commented):
Really good job Martin. The document page works fine and there are no performance problems at the moment. It works smooth (in the web view). 

The only thing i noticed is, that the Document page is checked in the side menu from the beginning. I think it would be better, if its checked after the applicant was on the page and pressed "Continue" for example, even if he doesn't want to upload a picture. So its still optional, but not check marked from the beginning.

frontend/src/app/forms/forms-routing.module.ts
MBuchalik: Hmmm... Tbh, I think it is fine if we leave it like it is for now since we will only deploy this as a "real" app (here, nobody is able to manually route to a random page). But we should keep this in mind and maybe think of some better solution later on. :)

MBuchalik:
> The only thing i noticed is, that the Document page is checked in the side menu from the beginning. I think it would be better, if its checked after the applicant was on the page and pressed "Continue" for example, even if he doesn't want to upload a picture. So its still optional, but not check marked from the beginning.

As we now have discussed in Slack, this needs to be discussed with the team. Maybe, we need another icon for that...

denoribi (approved):

PR #70, opened 2020-06-15
Privacy policy page added

Bharathravi1995:
This PR contains: 

- privacy policy  page changes 
- Rating page autosave changes
- Applicants page max-width changes

@vishwasanavatti  Can you review this?


frontend/src/app/rating/rating.page.ts
```diff
@@ -99,6 +99,30 @@ export class RatingPage implements OnDestroy, OnInit {
...
+          );
+          this.ratingForm.controls.applicantScore.controls.personalImpression.setValue(
+            applicantData.applicantScore.personalImpression
+          );
+          this.ratingForm.controls.applicantScore.controls.selfAssurance.setValue(
+            applicantData.applicantScore.selfAssurance
+          );
+          this.ratingForm.controls.impressionInfo.controls.impression.setValue(
+            applicantData.impressionInfo.impression
+          );
```
MBuchalik: I know you haven't requested a review from me. But one thing that I was wondering: Can't you just use `patchValue()` here? 

This would probably (careful, I haven't tested it!) look like this:

``` ts
if (applicantData) {
  this.ratingForm.patchValue(applicantData);
}
```

Bharathravi1995: Did not know such method exist. Will change. Thanks

vishwasanavatti (commented):
Checked it locally. It looks clean and good. 
Just one thing I noticed is I can just slide the "I have read" toggle and continue without even looking at what is there in the policy. Shouldn't it be policy has to be opened to toggle the agreement?

MBuchalik:
> Just one thing I noticed is I can just slide the "I have read" toggle and continue without even looking at what is there in the policy. Shouldn't it be policy has to be opened to toggle the agreement?

For now, I think it is fine to leave it like it is. We could of course refine this later on, but since the Industry Partners haven't provided any requirements regarding this, I would keep it as simple as it is now. 

(I haven't tried out the implementation. I just assume that it looks similar to the mockups.)

Bharathravi1995:
> Checked it locally. It looks clean and good.
> Just one thing I noticed is I can just slide the "I have read" toggle and continue without even looking at what is there in the policy. Shouldn't it be policy has to be opened to toggle the agreement?

Ya I also agree with your point. But we can iterate it over as this task is not complete until we add policy details.

vishwasanavatti (approved):

PR #71, opened 2020-06-15
Exit warning when leaving form pages

denoribi:
Implemented a CanDeactivate Guard to pretend accidentally leaving the form pages. 
Showing an alert message to handle leaving the Page correctly. 
Disabling swipe gesture on forms pages at iOS.
Upgraded ionic.
Making image viewer compatible with the Guard.

Thanks to @MBuchalik for helping me at this task :) Can you review this ? :)


frontend/src/app/forms/auth-guard.ts
```diff
@@ -0,0 +1,36 @@
...
+
+import { FormsPage } from './forms.page';
+import { FULL_IMAGE_VIEWER_PATH } from './router-constants';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class AuthGuard implements CanDeactivate<FormsPage> {
+  /**
+   * CanDeactivate method of the Guard to pretend going back to home page accidently without a warning
```
MBuchalik: There is a typo: "pretend" instead of "prevent". And "accidently" instead of "accidentally".

Also, I would prefer a wording like "... going back to the previous page ..." since it is not necessarily the home page the user might go back to.

frontend/src/app/forms/forms.page.ts
```diff
@@ -279,4 +319,63 @@ export class FormsPage implements OnInit, OnDestroy {
     }
     this.progressPercentage = validSteps / totalNumberOfRequiredSteps;
   }
+
+  /**
+   * Showing an alert message, when leaving a form page.
+   *
+   * returning a Promise<boolean>
+   * returning true, if user wants to leave the page.
+   * returning false, if user cancels leaving the page.
```
MBuchalik: My suggestion: Replace this with

```
@returns A Promise. This Promise will emit true if the user wants to leave the page. Otherwise, false is emitted.
```

Then you don't need this "returning, returning, returning".

frontend/src/app/forms/forms.page.ts
```diff
@@ -279,4 +319,63 @@ export class FormsPage implements OnInit, OnDestroy {
...
+  }
+
+  /**
+   * Deciding Function to handle leaving forms page, needed for the Guard
+   *
+   * @param hasSubmittedForm Boolean to check, if the applicant wants to leave the forms page by submitting,
+   * if applicant wants to submit, mayLeaveView() return true, so Guard will allow leaving the page.
+   *
+   * returning true = Guard will allow leaving the page.
+   * returning false = Guard will stop leaving the page.
```
MBuchalik: This method does not receive a param. So please remove the `@param` lines.

And I would maybe replace the lines starting with "returning" with sth like:

```
@returns true (either directly or emitted in the Observable) if the view may be left. Otherwise, false is returned (or emitted).
```

MBuchalik (commented):
Very good stuff 👍 I added a few comments regarding your comments :D

MBuchalik (approved):
Since there are only some commenting things you need to fix, I decided to approve this PR. Once you are done with the changes, you can directly merge it 👍

PR #72, opened 2020-06-15
Client Server Integration 

vishwasanavatti:
This PR handles integration of client and server. 

It involves upload page where number of applications being uploaded can be seen.
To test:  Please specify the path in Backend/ConfigProperties.java - absoluteFileSystemPath = "xx\\xx\\"; (set some local path). 
a) Error is shown when connection to server cannot be established.
b) When file path is not found.


dalex135 (approved):
Ran and checked. Works perfectly.

Backend/xx\xx\ab64f856-305a-4669-a1f1-3dbcd0d21b20.txt
```diff
@@ -0,0 +1 @@
...IXaTzRkgYpQ2DgcilK5GaRQgIPXrSYIyKUHnOKUDcSe1ADQxAxS57Hml8sdzTSuKAHAALnvTcEEUozjFKVJFAWDDKN2aZk5zThyMUjKBigAJyOlKozg07ggDFIRsoAdgZppDE03JNPbKqCRgmjYBhBHU0qqDyTQcEZ70g4OaBAzZ6Uq8gjNBPBOKVFGN3pQAzBowPWnsxb5qZnimFhxwcAdqaFJOMU4IcZBp24r2pAR5weKc3QGm96DwetMBSeoNAXIz3pCR0xk0h4HFArocY+M5pnTmlD+tITzTJuhQSM+9B4waQEZ5pOaAuOAJHFIaXLLyO9J1oATrS496SigQZNISad0oIPemS7jcmlGeaPak6UxB0602nnJppFCJYHg8UA4OaQcEd6CMk0xDsljSHrScjkUH1osO4u0YppGDTj04pCQfwoExpooAyeKMUyQBxQSTS0lADunFJnFLmigoemAcnmhsMc9qTOKXA25zzU+ZQhAz1oJ7UhGKBQAoyOaUnJyaSjpQMXBxntSUu44wOlKSPTmkAnIpdxxSUuMjIP4UFahkmgdcUYNKOvrSGDLgdaaMUrNnjtSAZoBijpx1p21hgEYpvSnbmIySaGNDtuzhhz7U0AkHFByVzmk5A9jSHcf82MZ4pnSnbwvQUgxjIPNA7gGK0h55ozk0pOaADHGaM5NJz0oAzQAoGOaBx0pRnPI470Zy3AwKAsGCckCl3HHNDNj5aTdjikMQZBzTtxZqQPj6Uo55pgAAJweKcrYznpTSPzpoGeKQ7jye1JyPegYH1pTwcigdwzkZpu4+tBbOaQ5BoFceuOtMLZo68ilC7h9KYhd5yOKcTvHJ6VGQRTlYKOlJoBM9BSu2TjORSdRngUBTjNMY7b8mc0n3RgikBxkUnbrQIUnt2pBmjHA9ac+c4FACIAT1pD1pOhzUnGM+tAheCvv6UzJFJ70E8e9Fhg2PxpuKKcM46ZFBIADBOaGG0fWg44xSMCp5oDQQ89BR1pQSOaMgn0pk2EII60uKVjke9N6UDAk0AZoXHcUuCW9KCdRuPSk7U4jBoPrTFYMjGMc0nXvS8YoGM0h2G0p5oIoIxTFYBTTTiMCm00JgB3ppODxTx6HpTSOaZDFzkUHG3HejtSjFAxpBAppp5JPFAQc5PSi5LGr1p2c9aZ1PFLk4pgmKACeaMAUm7ApM5oDQ/9k=
```
Bharathravi1995: For what this file is used?

frontend/src/app/app-routing.module.ts
```diff
@@ -29,6 +29,13 @@ const routes: Routes = [
         (m) => m.ApplicantsPageModule
       ),
   },
+  {
```
Bharathravi1995: I suggest to have this under applicants route like appliants/upload

frontend/src/app/applications-upload/applications-upload.page.ts
```diff
@@ -0,0 +1,193 @@
...
+    private toastController: ToastController
+  ) {}
+  /**
+   * The total application size.
+   */
+  totalSize = 0;
+  /**
+   * The application size that is being uploaded.
+   */
+  uploadSize = 1;
```
Bharathravi1995, line 32: According to the text this size corresponds to uploading count. So shouldn't it be 0 initially?

frontend/src/app/applications-upload/applications-upload.page.ts
```diff
@@ -0,0 +1,193 @@
...
+   */
+  url = 'http://localhost:9090';
+
+  /**
+   * In this method locally stored data are fetched and post request function is called for one application at time.
+   */
+  ngOnInit(): void {
+    this.storage
+      .getAllItems<FormsData>(this.storage.applicantDetailsDb)
+      .then(async (data) => {
```
Bharathravi1995, line 56: Why it has to be async and await?

frontend/src/app/applications-upload/applications-upload.page.ts
```diff
@@ -0,0 +1,193 @@
...
+        const applicantDetailsList = data.filter((x) => x.isRated === 1);
+        this.totalSize = applicantDetailsList.length;
+        for (const x of applicantDetailsList) {
```
Bharathravi1995: I suggest to use more meaningful variable name for x

frontend/src/app/applications-upload/applications-upload.page.ts
```diff
@@ -0,0 +1,193 @@
...
+          if (!this.isError) {
```
Bharathravi1995, line 60: Is this check needed? Let's say there are 100 records. And 1st API fails for some reason, then isError will become true. And rest of the API which were yet to be sent to the method won't be sent.

frontend/src/app/applications-upload/applications-upload.page.ts
```diff
@@ -0,0 +1,193 @@
...
+  /**
+   * In this method timeout of 1 second is speified to have a better readability of the application upload.
+   */
+  sleep(ms: number): Promise<void> {
+    return new Promise((resolve) => setTimeout(resolve, ms));
+  }
+  /**
+   * In this method a post request is made to the server.
+   */
+  sendPostRequest(inp: FormsData): Promise<void> {
```
Bharathravi1995, line 75: Why it needs to return promise? Is there any specific handling needed for promise?

frontend/src/app/applications-upload/applications-upload.page.ts
```diff
@@ -0,0 +1,193 @@
...
+    return this.sleep(1000).then(() => {
```
Bharathravi1995, line 76: Why there is need to delay in sending request?

frontend/src/app/applications-upload/applications-upload.page.ts
```diff
@@ -0,0 +1,193 @@
...
+      Object.entries(inp.keyCompetencies).map(([k, v]) => {
+        v.forEach((x: KeyCompetenciesEntry) => {
+          keyCompetence.push({ category: k, name: x.name, rating: x.rating });
+        });
+      });
+      /**
+       * education information are mapped according json format required by server.
+       */
+      Object.values(inp.educationInfo.educationInfoForm).map((v) => {
+        education.push(v);
```
Bharathravi1995: Can't we directly use the array from storage as there is no need of restructuring?

frontend/src/app/applications-upload/applications-upload.page.ts
```diff
@@ -0,0 +1,193 @@
...
+      });
+      /**
+       * picture of applicant is mapped according json format required by server.
+       */
+      images.push({
```
Bharathravi1995, line 95: Need to add documents pics as well.

frontend/src/app/applications-upload/applications-upload.page.ts
```diff
@@ -0,0 +1,193 @@
...
+        .subscribe(
+          (response) => {
+            if (response.status === 201) {
+              this.storage.deleteItem(this.storage.applicantDetailsDb, inp.id);
+              this.storage.deleteItem(this.storage.applicantRatingsDb, inp.id);
+              if (this.uploadSize === this.totalSize) {
+                this.isSuccess = true;
+                this.completionAlert();
+              } else {
+                this.uploadSize++;
```
Bharathravi1995, line 140: In html text it says uploading so shouldn't be incremented after sending API? Now it corresponds to uploaded count.

frontend/src/app/applications-upload/applications-upload.page.ts
```diff
@@ -0,0 +1,193 @@
...
+              }
+            }
+          },
+          (error) => {
+            this.isError = true;
+            if (error.status === 0) {
```
Bharathravi1995: There is no HTTP status code 0. 
1.The Client cannot connect to the server
2) The Client cannot receive the response within the timeout period
3) The Request was "stopped(aborted)" by the Client.
These might be reasons for 0. May be what we can have is Specific error msg in if else if then a Generic message in else for cases like unhandled error codes including 0.

Bharathravi1995 (commented):
Good job @vishwasanavatti . It was really huge task. I am glad that you cracked it on time. I checked the frontend part. It works fine.  Have added few suggestions and doubts I had as comments. Kindly check it.

Backend/xx\xx\ab64f856-305a-4669-a1f1-3dbcd0d21b20.txt
vishwasanavatti: This is not necessary. This is actually the image file which gets saved if we dont specify the file path. will remove this.

frontend/src/app/applications-upload/applications-upload.page.ts
vishwasanavatti: I feel we always start upload from 1 file and not 0 file so it is 1.

frontend/src/app/applications-upload/applications-upload.page.ts
vishwasanavatti: ya that can be done. will change

frontend/src/app/applications-upload/applications-upload.page.ts
vishwasanavatti: To wait for the response from the postRequest task to check if there is any error. if there is error it just do not process further applications.

frontend/src/app/applications-upload/applications-upload.page.ts
vishwasanavatti: I tried it but it didn't seem to work.

frontend/src/app/applications-upload/applications-upload.page.ts
vishwasanavatti: ya but the document page is implemented in the current sprint. I can add it now as it has been merged or I can add in the next sprint.

frontend/src/app/applications-upload/applications-upload.page.ts
vishwasanavatti: It actually doing the same. Only on success of previous API, it is being incremented.

frontend/src/app/applications-upload/applications-upload.page.ts
vishwasanavatti: When server is not up. I get ERR_CONNECTION_REFUSED with status as 0. I will just handle this in else part then.

frontend/src/app/applications-upload/applications-upload.page.ts
vishwasanavatti: This is required because of readability purpose or else the page is of no use. The process happens in no time(in ms) and we wont even get to know what happened. If required we can reduce the time for large number of records by adding condition to set time value.

frontend/src/app/applications-upload/applications-upload.page.ts
vishwasanavatti: it is because of sleep function inside it returns promise.

frontend/src/app/applications-upload/applications-upload.page.ts
vishwasanavatti: > I tried it but it didn't seem to work.

My bad it works. I had tried in a wrong way.

frontend/src/app/applications-upload/applications-upload.page.ts
vishwasanavatti: Added documents to images as well.

Nasser-Eddin-Nasser (approved):
@vishwasanavatti showed me a demo, and we did testes on it. The code quality is good as well. 
In my opinion, it can be merged  :)

frontend/src/app/app-routing.module.ts
vishwasanavatti: I guess even applicants need to be added under home, so can we just keep this open and do everything in next sprint?

frontend/src/app/applications-upload/applications-upload.page.ts
vishwasanavatti: ya thats true but now the error is more genera. If one application fail, all the corresponding applications fail. We can change this with more specific errors in the coming sprint.

MBuchalik:
I just had a very brief look at the code. In general, a really good start 👍 What I didn't see here: When you go back (either by clicking "cancel", by swiping back, clicking the hardware back button, ...), will the uploading be actually cancelled?

Let's say there are 10 applications in the queue and you navigate back after the second one. Will the other 8 still be uploaded? From my point of view, the uploading should be cancelled in this case.

One trick for that: If you unsubscribe from an HttpClient's Observable, then the request will also be cancelled. Don't get me wrong on this one: Usually, you don't need to unsubscribe from HttpClient Observables, since they destroy temselves automatically. But for this special use case, you could maybe implement it like that. Or, alternatively, just stop the queue (so the current one will still be uploaded, but then the upload queue stops).


Another thing: Did I see it correctly that you don't wait for one request to complete, but rather have a fixed timeout of one second? From my point of view, it should rather be done like so:

Send Application 1 --> Wait for completion --> Send Application 2 --> Wait for completion --> ...

But if I am not mistaken, you simply create a new request every second. But what if the connection is slower than that?

Bharathravi1995:
> Let's say there are 10 applications in the queue and you navigate back after the second one. Will the 

> other 8 still be uploaded? From my point of view, the uploading should be cancelled in this case

Ya it has to be aborted if there is any pending request or unsubscribed u order to not send requests further. It is mentioned in the issue.

> Send Application 1 --> Wait for completion --> Send Application 2 --> Wait for completion --> ...

Should api be sent one by one? From my point of view we can have a threshold, let's say in chrome maxm of 6 requests per host  can be processed at a time. So we can just allow 6 at a time when a requests resolves can add one . This way the requests will be processed parallelly upto a threshold.  On the other hand we can just directly trigger api without setTimeOut or waiting for an API to be resolved. Browser will manage the network queue. But this has an adverse effect of blocking and bombarding with requests. But I am not sure how hybrid apps handles API, and maximum request it process parallelly.

MBuchalik:
> Should api be sent one by one?

I just thought it would be more simple to build it like this. Of course, we can just generate all requests at once and fill the browser's queue with that. Or generate n requests. But I guess that sending only one request at a time is less error-prone (for instance, the cancellation logic is more simple since you really only need to cancel one request). And it is also easier for the user to understand.  Also: I don't think that sending one request at a time is too slow either. At least for now, this should be fine.

Btw: I am not sure if the limit of simultaneous requests is always 6. It could be a different number (less or more), so we would need to find a way to detect this. Or hardcode this to a small number like 3 or 4.

Summary: I think that sending one request at a time is the most simple solution that is also easily testable and less error-prone compared to other solutions. Only if the Industry Partners say they want maximum speed, then, from my point of view, we should think about parallelization.

That is at least my point of view. :)

Bharathravi1995:
I completely agree it is simple. But my point is why can't we use the advantage of parallalization when it is available(But for dat we should find a suitable threshold) . It is a race between simplicity vs speed. Ok let's first implement one by one i. Then will think of paralization.

MBuchalik:
I am not completely against building a parallel version. But my idea is: First build a simple non-parallel one (with support for cancelling, good error handling etc.). And then, if we have enough time left, build a parallel one. :)

PR #77, opened 2020-06-19
Create Admin Account

Nasser-Eddin-Nasser:


HR-Monitor/hr-monitor.properties
```diff
@@ -0,0 +1,4 @@
+path.variable.kotlin_bundled=C\:\\Program Files\\JetBrains\\IntelliJ IDEA Community Edition 2019.2.3\\plugins\\Kotlin\\kotlinc
+path.variable.maven_repository=C\:\\Users\\EDDIN\\.m2\\repository
+jdk.home.11.0.4=C\:/Program Files/JetBrains/IntelliJ IDEA Community Edition 2019.2.3/jbr
+javac2.instrumentation.includeJavaRuntime=false
```
vishwasanavatti, line 4: what is this file for? why are paths hard coded?

vishwasanavatti (changes requested):
Tested locally. I could able to create an account but was not able to login. Could you check the issue?

HR-Monitor/hr-monitor.properties
Nasser-Eddin-Nasser: It's for the Ant builder, it should be in git ignore I guess :)

vishwasanavatti (approved):
Tested again after changes and its works fine. Good work @Nasser-Eddin-Nasser, its perfectly done. All the basic functionalities are included. Can be merged.

dalex135 (approved):
Ran and tested. Well done. Quality code. :)
Was easy to merge with the US 42.

MBuchalik:
Edit: Sorry - I missed someting. I initially thought the passwords would be stored in cleartext :D (Or that some sort of encryption would be used instead of a hash.)

Would be cool if we used sth like a salted hash, but we can add this in the future as well.

PR #79, opened 2020-06-20
Native iOS build

MBuchalik:
I have made a native iOS build. This PR includes the files that are needed for this. Also, I have fixed many (performance) issues.

I have added a few screenshots into our shared folder so that you can check how the app looks like. Also, if you like, you could also make an XCode build.

Tbh, I don't think there is much to review here.


frontend/package.json
```diff
@@ -26,6 +26,7 @@
     "@angular/platform-browser-dynamic": "~9.1.9",
     "@angular/router": "~9.1.9",
     "@capacitor/core": "2.2.0",
+    "@capacitor/ios": "^2.2.0",
     "@ionic-native/core": "^5.26.0",
```
denoribi, line 30: Don't forget to add this in the BOM ;)

denoribi (approved):
Looks fine for me, I will merge it then !

frontend/package.json
MBuchalik: Thanks for the reminder! 👍

PR #81, opened 2020-06-23
Internationalisation in English and German languages supported

Bharathravi1995:
**This PR contains following changes:**

- Language picker added (Items in language picker are not translated intentionally because the language has to be in its own script)

- Translation of strings in HTML, TS, Placeholders, Toast messages, Default ionic text e.t.c

**Yet to do:**

- Key competencies page languages are not internationalised as it involves changing the structure of the component. Currently the display names and values in the storage are same. Need to separate display name and values. After discussion with the person who implemented it, I will change it within next sprint.

- Other fields like  Programming languages and Frameworks etc  are also not translated, as decision needs to be taken because Frameworks and technologies remain the same irrespective of  language. Since for some fields in key competencies translation needed and others not, a common pattern has to be implemented which I will do it within next sprint.

- Once changes are merged have to update BOM.


denoribi (approved):
So except the languages fields on key competencies everything looks fine for me.  Good job :)

frontend/src/app/home/language-picker/language-picker.component.ts
```diff
@@ -10,16 +15,28 @@ import { StorageHandlerService } from '../../services/storage-handler.service';
   styleUrls: ['./language-picker.component.scss'],
 })
 export class LanguagePickerComponent {
+  /**
+   * Holds the supoorted languages code, display name and chosen flag.
+   */
   supportedLanguages = [
     { localeCode: 'en', displayName: 'English', isChosen: false },
     { localeCode: 'de', displayName: 'Duetsch', isChosen: false },
```
denoribi: Here is a typo, "Deutsch" ;)

denoribi (approved):

frontend/src/app/home/language-picker/language-picker.component.ts
Bharathravi1995: Oh yeah thanks for notifying. Will fix it.

PR #82, opened 2020-06-23
Language header added in language picker

Bharathravi1995:


vishwasanavatti (approved):
It looks great. Checked locally and everything translates. Good job, It can be merged.

PR #84, opened 2020-06-24
add  architecture diagram

Nasser-Eddin-Nasser:


dalex135 (approved):
Checked it. Architecture is correctly drawn.

PR #85, opened 2020-06-24
Improve forms types

MBuchalik:
As discussed today, I implemented some small improvements to the forms types. This is all still an "experiment", so it might happen that you experience some issues with that. The most notably things: 
1. FormControl now accepts booleans
2. Our forms now use `.value` instead of `.getRawValue()` when storing the data in Ionic Storage

Also, I have added 2 "TODOs" to the application upload component file. We should try to fix/implement them soon.


MBuchalik:
Sorry for this small force-pushing. I accidentally committed a line that was only used for dev purposes.

denoribi (approved):
Looks good for me :) will merge it

PR #86, opened 2020-06-25
Add Business Skills to Key Competencies, add various improvements

MBuchalik:
I did a few things:
- Added "Business Skills" to our Key Competencies page
- Also extended the server to handle this (important: I did nothing with the HR monitor! Not sure if something needs to be adjusted here)
- Added various changes to both language files
- Added some small additional improvements, e.g. regarding the rating page


dalex135 (approved):
Ran and checked both Backend and frontend. Both work perfectly.

PR #91, opened 2020-06-29
64-check-steps

denoribi:
Implementation of the possibility to skip steps.


frontend/src/app/applications-upload/applications-upload.page.ts
```diff
@@ -108,10 +102,10 @@ export class ApplicationsUploadPage implements OnInit {
        * picture of applicant and documents is mapped according json format required by server.
        */
       images.push({
-        content: inp.profilePicture.pictureBase64,
+        content: inp.profilePicture?.pictureBase64,
```
MBuchalik: I am not sure if this is how the server expects the data. Shouldn't everything here rather be wrapped in an if statement? Otherwise, the server will receive a profile picture with no content.

frontend/src/assets/i18n/de.json
```diff
@@ -194,5 +194,11 @@
     "sideMenu": {
         "progress": "Fortschritt",
         "description": "Wenn Du zu einem vorherigen Schritt wechseln willst, dann klicke einfach auf den entsprechenden Eintrag in der Liste!"
+    },
+
+    "skipStep":{
+        "skip": "Überspringen",
+        "skipMessage": "Möchtest du diesen Schritt wirklich überspringen? Bitte überspringe diesen Schritt nur dann, wenn die hier benötigten Informationen Teil der Dokumente sind, welche Du als Bilder hochgeladen hast.",
```
MBuchalik: Please write "Du" with a capital "D"

frontend/src/theme/custom.scss
```diff
@@ -90,13 +92,13 @@ ion-item.item-has-focus {
 */
 
 .main-container {
-  width: 100%;
```
MBuchalik, line 93: Why have you removed this line?

MBuchalik (approved):
Edit: Sorry, I missed something. So I had to take back the approval.

Have you actually tested it with a disabled field? I got the feeling that you have missed to handle the case where these fields don't exist: https://github.com/Nasser-Eddin-Nasser/Digitact/blob/2d77bd95e5554db92eeb490a401b68317200e6bd/frontend/src/app/applications-upload/applications-upload.page.ts#L123-L125
Also: What happens if no KeyCompetencies are available? https://github.com/Nasser-Eddin-Nasser/Digitact/blob/2d77bd95e5554db92eeb490a401b68317200e6bd/frontend/src/app/applications-upload/applications-upload.page.ts#L96


My suggestion: "Skip" every field that can currently be skipped. And then try out what happens if you send it to the server.

frontend/src/theme/custom.scss
denoribi: Maybe a copy/paste error, added this line again ;)

frontend/src/app/applications-upload/applications-upload.page.ts
denoribi: actually the profile picture is not part of the skipped steps, so i just deleted the "?". So it works for the moment, because there will be always a profile picture atm

Nasser-Eddin-Nasser (commented):
I reviewed it with Denis on a video call, and I get through the commits. It works perfectly but the code structure can be a little bit better, but for this sprint, I think it's fine and can be merged :)

MBuchalik (approved):
Looks good 👍 

From my point of view, you can merge it.

Nasser-Eddin-Nasser (commented):
Merge conflicts must be resolved, then it can be merged :)

Issue #41, opened 2020-05-26
Ionic storage related issues


Bharathravi1995:

Bharathravi1995:
- [x] Async await support in functions

- [x] Error handling needs to be done

- [x] Backup mechanism to store in sqlite

- [x] Storing images and files in a better way

- [x] Creating multiple store instance support

PR #92, opened 2020-06-30
US 67: SQLite support added inorder to prevent data loss

Bharathravi1995:
**Done:**

- Now the local forage library prioritise the driver based on the availability and our preference in drive order.
- SQLite added as first priority.
- In web view Indexed db will top priority.
- Cordova sqlite plugin was added as npm package and synced with capacitor to build it.
- CRUD methods will be returning promises of local forage library directly instead of using ionic storage.
- General error handling done.

It has the  fix for https://github.com/Nasser-Eddin-Nasser/Digitact/issues/41

**Yet to do:**

- Updating BOM
- Updating issue

**Need to fix:**

In the applicants page sudden switching between No item and Item listing of applicants happens because it takes time to resolve promises. Need to add loading state until promise resolve to fix this issue.


vishwasanavatti (approved):
Good job as you were able to do this in time. 
Approving it as I and Bharathwaj were able to test in local system by setting up ios simulator. It works and data are being stored in SQLite.

PR #94, opened 2020-06-30
Applicant Info

dalex135:
A new PR raised as there was an issue in Work experience not showing in my previous branch.


vishwasanavatti (approved):
Tested locally and all the data are shown in the HR monitor. But few alignment changes and testing document with real document pictures. Its ok for now so approving for merge.

PR #98, opened 2020-07-03
Dark Mode and Android support

MBuchalik:
This PR:
- improves the Dark Mode design on iOS
- improves the UI on Android
- makes it possible to compile the app for Android
- improves the Dark Mode design on Android
- fixes lots of small issues


denoribi (approved):
I just had a quick look at the code without beeing able to check it live, but it looks good for me. I will merge this before we get new merge conflicts or something like this.

good job Martin ;)

PR #100, opened 2020-07-05
HR applicant status and key components

Nasser-Eddin-Nasser:


dalex135 (approved):
Checked and ran. Works really good.

PR #101, opened 2020-07-06
75 profile picture

denoribi:
Profile Picture is now optional. 
- If there is a picture it will be send to the server, if there is no profile picture, we send and empty array or just the documents pictures without the profile picture. 
- Backend/HR-Monitor changes to accept no profile picture too.

### DIFF
--- a/HR-Monitor/src/Controller/ApplicantInfoController.java
+++ b/HR-Monitor/src/Controller/ApplicantInfoController.java
@@ -370,23 +370,28 @@ public class ApplicantInfoController {
     }
 
     private void setProfPic(List<AppImage> images) {
-        AppImage profImage =
-                images.stream()
-                        .filter(x -> x.getType().equals(ImageType.profilePic))
-                        .findFirst()
-                        .get();
-        if (profImage != null) {
-            try {
-                if (profImage.getContent() == null)
-                    Connector.sendGetHttp(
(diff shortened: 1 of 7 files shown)


HR-Monitor/src/Controller/ApplicantInfoController.java
```diff
@@ -370,23 +370,28 @@ private void getImages(ImageType it) {
...
+        AppImage profImage = null;
+        boolean hasProfilePicture = false;
+        for (AppImage img : images) {
+            if (img.getType() == ImageType.profilePic) {
+                hasProfilePicture = true;
+                profImage = img;
+            }
+        }
+        if (hasProfilePicture) {
+            if (profImage != null) {
```
vishwasanavatti: why hasProfilePicture variable is required? it does not make any difference here. variable profImage is enough and it doing the same thing as hasProfilePicture.

vishwasanavatti (changes requested):
Checked locally, everything works perfect. Just you need to remove the unnecessary variable.

HR-Monitor/src/Controller/ApplicantInfoController.java
denoribi: Yep i see it, that mistake came from refactoring this. i will correct it

vishwasanavatti (approved):

PR #107, opened 2020-07-08
config Assertion and Cleanup finction

Nasser-Eddin-Nasser:


vishwasanavatti (approved):
checked locally it works fine. ok to merge

PR #113, opened 2020-07-12
Filtering of Applicants in HRM

dalex135:
Please test this really well. and let me know if there is any bug.

### DIFF
--- a/Backend/src/main/java/Digitact/Backend/ConfigProperties.java
+++ b/Backend/src/main/java/Digitact/Backend/ConfigProperties.java
@@ -7,7 +7,7 @@ import org.springframework.context.annotation.Configuration;
 @Configuration
 public class ConfigProperties {
     public static final int serverPort = 9090;
-    public static final String absoluteFileSystemPath = "D:\\Masters\\2nd Semester\\AMOS Project\\FileSystem\\";
+    public static final String absoluteFileSystemPath = "";
     public static final String BLOCKFORMAT = ".txt";
     public static final long BLOCKSIZE = 1024 * 1024; // 1 MiB
     public static final int Max_Repetition_Try = 5;
--- a/Backend/src/main/java/Digitact/Backend/ConfigProperties.java
+++ b/Backend/src/main/java/Digitact/Backend/ConfigProperties.java
@@ -7,7 +7,7 @@ import org.springframework.context.annotation.Configuration;
 @Configuration
 public class ConfigProperties {
     public static final int serverPort = 9090;
-    public static final String absoluteFileSystemPath = "";
+    public static final String absoluteFileSystemPath = "D:\\Masters\\2nd Semester\\AMOS Project\\FileSystem\\";
     public static final String BLOCKFORMAT = ".txt";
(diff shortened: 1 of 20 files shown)


Backend/src/main/java/Digitact/Backend/ConfigProperties.java
```diff
@@ -7,7 +7,7 @@
 @Configuration
 public class ConfigProperties {
     public static final int serverPort = 9090;
-    public static final String absoluteFileSystemPath = "";
+    public static final String absoluteFileSystemPath = "D:\\Masters\\2nd Semester\\AMOS Project\\FileSystem\\";
```
vishwasanavatti: Make sure you dont push this change.

HR-Monitor/src/Main/Configuration.java
```diff
@@ -6,7 +6,7 @@
 public class Configuration {
     public static final String BES_URI = "http://localhost:9090/api/HRController/";
 
-    public static final String absoluteFileSystemPath = "";
+    public static final String absoluteFileSystemPath = "D:\\\\Masters\\\\2nd Semester\\\\AMOS Project\\\\FileSystem\\\\";
```
vishwasanavatti: same here! dont merge this change.

HR-Monitor/src/Util/Dictionary/IDictionary.java
```diff
@@ -0,0 +1,39 @@
...
+        if (LANG.equals(Language.English)) {
+            if (english.contains("TITLE-")) // GUI title
+            return english.replace("TITLE-", "");
+            return english;
+        }
+
+        return getGerman(dictionary, english);
+    }
+
+    private static String getGerman(IDictionary dictionary, String english) {
```
vishwasanavatti: shouldn't the access modifier be public?

vishwasanavatti (commented):
Great job, the filter thing just works perfect. Also the translation between German and English works great as well. I found alignment issues at 2 places due to translation and also have a question. 
Filter works fine but is it good idea to call a query at each select or entry in the field? it just keeps calling query and affects the performance I fell.  I prefer to have a search button and on click of it a query will be called only once and data will be fetched.

Backend/src/main/java/Digitact/Backend/ConfigProperties.java
dalex135: Yep. these will be corrected in the last commit when I am about to merge.

dalex135:
@vishwasanavatti I corrected the DB calls for each and the German translations. 
@denoribi Pls check for the German strings here. If check here that's enough. This PR has all the changes regarding to translations.

denoribi (approved):
Translations look correct to me :)
Good job guys ;)

Nasser-Eddin-Nasser:
I ran the code locally and checked each commit. It works perfectly, just a small issue, by german version the position and industry still in the English version. 
image

vishwasanavatti (approved):
Checked locally and it works fine. ok to merge.

PR #114, opened 2020-07-13
Dalex135 hr filtering

Nasser-Eddin-Nasser:


dalex135 (approved):
Ran and tested the code. Works perfectly.

PR #116, opened 2020-07-14
Api protection

vishwasanavatti:
This PR consist of login page and token creation and has following features
1) A login page comes up when device starts for the first time and will be navigated again to login page only when the user token validity expires.
2) on login, A device token will be created which will be stored in client local storage and in DB permanently.
3) A user token will be created and stored in storage as well with expiration of 15 days. So after 15 days, user has to login again with same or different credentials as in HR monitor.


Bharathravi1995 (approved):
Checked the changes. It works fine. LGTM. Can merge it.

dalex135 (approved):
Tested the code with Vishwas. Job well-done guys. Work as intended.

PR #119, opened 2020-07-15
HR-Monitor Executable 

Nasser-Eddin-Nasser:


dalex135 (approved):
The bug in authentification fixed. Ran and tested. Great job!

PR #122, opened 2020-07-17
81 skip button update

denoribi:
- Skip allert message now shown only once
- Skip allert message changed to be more general
- Work experience page made skippable 
- (Additonal Information page should be optional, clarified with the POs)


frontend/src/app/forms/forms.page.ts
```diff
@@ -164,6 +164,11 @@ export class FormsPage
    */
   private hasSubmittedForm = false;
 
+  /**
+   * If a skip is stepped, skipAlertShowed = true, if no stepp is skipped till now, it returns false.
```
MBuchalik: Uhm, I think your comment message is a bit broken :D 

From my point of view, you could maybe summarize it like so:
"Has the user ever skipped a step?"

frontend/src/app/forms/forms.page.ts
```diff
@@ -164,6 +164,11 @@ export class FormsPage
...
+   */
+  private skipAlertShowed = false;
```
MBuchalik: The past tense is "shown" ;)

frontend/src/assets/i18n/en.json
```diff
@@ -220,7 +220,7 @@
 
     "skipStep":{
         "skip": "Skip",
-        "skipMessage": "Do you really want to skip this step? Please only do this if the data asked for in this step are part of the documents you took a photo of.",
+        "skipMessage": "If you skip this page, no information will be transferred. If possible, please provide all information at your hand - uploading a document is one way to do so.",
```
MBuchalik: Should we say "If you skip this page" or "If you skip this step"? Not sure...

frontend/src/app/forms/model/steps.model.ts
```diff
@@ -51,7 +51,7 @@ export const ApplicationStepsConfig: ApplicationStepsConfig = {
   [ApplicationStep.WorkExperience]: {
     useForProgressCalculation: true,
     formItemName: 'workExperienceInfo',
-    maySkipStep: false,
+    maySkipStep: true,
```
MBuchalik, line 54: (This comment is not really related to this line...)

When you skip the Work Experience step, then in the right menu, it will still not be "checked". The reason: There is a bug in `forms/side-menu/side-menu.component.html`. Normally, for every menu item, we define "isDisabled". However, someone (probably me :D ) has forgotten to define this for the Work Experience step: 

grafik

So, could you add the missing definition for "isDisabled" to the side menu html? Then, when skipping the Work Experience step, the checkmark should hopefully appear.

MBuchalik (commented):
Nice! Skipping steps now feels much "smoother". 👍

frontend/src/app/forms/forms.page.ts
denoribi: Changed it ;)

frontend/src/app/forms/forms.page.ts
denoribi: Yeah, you are right :)

frontend/src/assets/i18n/en.json
denoribi: Step is better, In German it was "Schritt" too, so i changed it in english too :)

frontend/src/app/forms/model/steps.model.ts
denoribi: Yes, this solved the problem :)

MBuchalik (approved):
LGTM 👍

PR #125, opened 2020-07-19
Diagrams for the BES and HRM

Nasser-Eddin-Nasser:


vishwasanavatti (approved):
Great job. Its is really good.

PR #129, opened 2020-07-19
Update work experience Screenshots

denoribi:
Updated the work experience Screenshots


MBuchalik (commented):
You have created the screenshots for Android. Since we are mostly using the iOS style, could you please also do so? (To be more precise: In Chrome, we use the "iPhone 6/7/8" UA and then use the "Capture screenshot" function you see when clicking on the three dots.)

MBuchalik (approved):

MBuchalik:
I have slightly changed the screenshots, hope that was ok for you.

